PR #12, opened 2019-06-04
rework build

jedwards4b:
Logic was moved from here to the cime Makefile.  Added pop ocn to Bcase runcmp


mvertens:
This change to CMEPS is dependent on a change to the CIME Makefile - so this dependency should be noted in the PR summary.

PR #27, opened 2019-07-10
changes needed for mom6 mct/nuopc validation

mvertens:
Change Description: changes needed for mom6 mct/nuopc validation

This PR adds the following:
- attributes of name and units to mediator history files
   these attributes now come from querying the NUOPC field dictionary and as a results the routines
   `shr_nuopc_fldList_AddMetadata` and `shr_nuopc_fldList_GetMeta` in `esmFlds` are not longer needed
- change  in `fd.yaml` of the alias names for the swpen coming from cice (these were duplicates and now have new names - as a result cice must also have a new hash)
- addition of new cart3d atm->ocn and atm->ice mapping that is triggered by the new config option `mapuv_with_cart3d`. This is currently to to .false. by default - but will be turned on as default in an upcoming tag.
- new run sequence for C and G compsets to match mct run sequence
- set of coldair_outbreak to .false. for T62 and TL319 datm forcing 

The above changes answers for DTESTS, C and G compsets.
Note that MOM restarts are now working!

Testing Completed: testlist_drv.xml - new baselines created are in jul09intel19/
(Minimum testing: CIME_DRIVER=nuopc scripts_regression_tests.py)

Externals coming soon!

CIME HASH:

### DIFF
--- a/mediator/med.F90
+++ b/mediator/med.F90
@@ -22,6 +22,7 @@ module MED
   use shr_nuopc_methods_mod , only : FB_Reset           => shr_nuopc_methods_FB_Reset
   use shr_nuopc_methods_mod , only : FB_Copy            => shr_nuopc_methods_FB_Copy
   use shr_nuopc_methods_mod , only : FB_FldChk          => shr_nuopc_methods_FB_FldChk
+  use shr_nuopc_methods_mod , only : FB_diagnose        => shr_nuopc_methods_FB_diagnose
   use shr_nuopc_methods_mod , only : clock_timeprint    => shr_nuopc_methods_clock_timeprint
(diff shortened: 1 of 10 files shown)


mediator/med.F90
```diff
@@ -1767,6 +1768,12 @@ subroutine DataInitialize(gcomp, rc)
     call med_fraction_set(gcomp,rc=rc)
     if (ChkErr(rc,__LINE__,u_FILE_u)) return
 
+    ! should this be added here?
```
jedwards4b: should this comment be removed?

mvertens:
not sure - that's why I've kept it in there for now.

Issue #31, opened 2019-09-03
Mediator cannot write history file in the first coupling time step


alperaltuntas:
This is because the mediator initializes history file alarm within the first coupling timestep, but time must advance for an alarm to be turned on after initialization, so the alarm cannot be turned on until at least the second coupling timestep. I am working on a fix for this and will follow up with a PR.

PR #32, opened 2019-09-03
Fixes history file write in first cpl timestep

alperaltuntas:
Change Description: Currently, history file alarm is initialized within the first coupling timestep. That means, history files cannot be written until the second timestep because for alarms to turn on, time must advance. With this PR, history alarm gets initialized before run starts. This change fixes issue #31 

Testing Completed: CIME_DRIVER=nuopc scripts_regression_tests.py

CIME HASH: 2b02a76de27e1d9abefcfcb773a69ec58e437e00


alperaltuntas:
bringing this to @mvertens's attention.

PR #39, opened 2019-11-04
changes to get CMEPS to work with both NEMS and CESM

mvertens:
Change Description: new directory structure and files to CMEPS to work with both NEMS and CESM

Testing Completed: testlist_drv.xml - new nuopc baselines created for nov03
    all tests were bit-for-bit except for namelist differences due to cice update
     nuopc scripts_regression_tests.py was not run - since this did not touch any code used by this test

CIME HASH: 0bd9a353e
app_cesm-cmeps HASH: 7aa91e22  
    (updates to cime, clm and cice hashes)


mediator/esmFldsExchange.F90
```diff
@@ -2118,7 +2118,7 @@ subroutine esmFldsExchange(gcomp, phase, rc)
     !-----------------------------------------------------------------------------
     ! TODO: add this
     ! if (carma_flds /= ' ') then
-    !    do n = 1,shr_string_listGetNum(carma_flds)
+    !    do n = 1,)number_of_fields in carm_flds)
```
jedwards4b: In comments but wrong as well - remove?

mediator/med_phases_prep_atm_mod.F90
```diff
@@ -4,6 +4,31 @@ module med_phases_prep_atm_mod
   ! Mediator phases for preparing atm export from mediator
   !-----------------------------------------------------------------------------
 
+  use med_kind_mod          , only : CX=>SHR_KIND_CX, CS=>SHR_KIND_CS, CL=>SHR_KIND_CL, R8=>SHR_KIND_R8
```
jedwards4b: why was this done?  It's better inside the subroutine.

mediator/med_phases_prep_ice_mod.F90
```diff
@@ -4,6 +4,24 @@ module med_phases_prep_ice_mod
   ! Mediator phases for preparing ice export from mediator
   !-----------------------------------------------------------------------------
 
+  use med_kind_mod          , only : CX=>SHR_KIND_CX, CS=>SHR_KIND_CS, CL=>SHR_KIND_CL, R8=>SHR_KIND_R8
```
jedwards4b: Why?

mediator/med_phases_prep_wav_mod.F90
```diff
@@ -4,6 +4,18 @@ module med_phases_prep_wav_mod
   ! Mediator phases for preparing wav export from mediator
   !-----------------------------------------------------------------------------
 
+  use med_kind_mod          , only : CX=>SHR_KIND_CX, CS=>SHR_KIND_CS, CL=>SHR_KIND_CL, R8=>SHR_KIND_R8
```
jedwards4b: these don't need to be globals

nems/util/shr_abort_mod.F90
```diff
@@ -0,0 +1,164 @@
+module shr_abort_mod
+  ! This module defines procedures that can be used to abort the model cleanly in a
```
jedwards4b: I notice above that you replaced shr_abort_mod with an ESMF abort call - do you still need this?

nems/util/shr_kind_mod.F90
```diff
@@ -0,0 +1,19 @@
+MODULE shr_kind_mod
```
jedwards4b: Why is this needed?  Thought you replaced it with med_kinds_mod?

mvertens:
@jedwards4b - its much easier to change things in one place then in
multiple places in the subroutine.

mediator/esmFldsExchange.F90
mvertens: I can remove this - but its a place holder to reference what is needed. I can add a comment as well as opposed to pseudo-code. Would that be better?

mediator/med_phases_prep_atm_mod.F90
mvertens: Actually I disagree. For the kind - it should be a module level use statement since it could apply to any subroutine that uses it.

mediator/med_phases_prep_ice_mod.F90
mvertens: Same answer as before.

mediator/med_phases_prep_wav_mod.F90
mvertens: I disagree.  I believe that the kind references should be global since they only then need to be defined in one place.

nems/util/shr_abort_mod.F90
mvertens: You still need this for for util - since it specifically references cime/src/share/util routines that use shr_kind_mod. We could rewrite those routines - but I did not take that on right now.

nems/util/shr_kind_mod.F90
mvertens: see above answer.

PR #42, opened 2020-01-03
changes to get aux_clm tests to work

mvertens:
Change Description: these are changes that were need to get the aux_clm tests to work.

Summary:

- The primary source of these changes was to enable the running of the aux_clm test suite. Howver, there is a still a problem with running any of the aux_clm tests with threading.
-  A much more flexible generation of nuopc run sequences was implemented. Almost all of the compsets now use runseq_general.py to generate the run sequences. As a result of this new run sequence, rof coupling times are now set to what would be expected from running cpl7 and not hard-wired to be the same as atm and lnd coupling times.
- A new xml variable `CREATE_ESMF_PET_FILES` was introduced to separate the creation of PET files from the DEBUG flag. By default this is false - but can now be activated for both debug and non-debug runs.
 
Testing Completed: testlist_drv.xml on cheyenne

- results were compared with the nov08 baseline and new jan05 baselines were created
- the testlist_drv.xml has been modified to include more realistic I2000 tests.
 - In the BMOM compsets, the default output for cam and clm history files is now double precision - this will cause the history file compares for clm and cam to be non-bfb (when the output was set back to single precision, the results were verified to be bfb with nov08)
- In the A compsets, the tests are non-bfb in the cpl history for just the ocean TS field.
  SMS_Vnuopc_Ln11_D.f19_g17_rx1.A.cheyenne_intel
  ERS_Vnuopc_Ln11_D.f19_g17_rx1.A.cheyenne_intel
  This is due to the change in the output sequence for the coupler history file. The new run sequence is correct.
- The following test failures from PR#40 was fixed:
  - ERS_Vnuopc_Ln11.f19_f19_mg17.I2000Clm50SpGs.cheyenne_intel.clm-default COMPARE_base_rest
  - ERS_Ly3_Vnuopc.f09_g17_gl4.T1850G.cheyenne_intel RUN 
However the fix caused the baseline comparison with the nov08 baseline to fail.
- The remaining test problems that were in #PR40 ia still there.

**CIME HASH**: abd7a3a7d (this is also going to have a PR and be updated once that PR is accepted)
**app_cesm-cmeps hash**: 0b8a58f


mediator/med.F90
```diff
@@ -2010,7 +2015,9 @@ subroutine SetRunClock(gcomp, rc)
 
     if (first_time) then
 
+       ! --------------------
```
jedwards4b: These decorations do nothing to add to the readability of the code in my opinion.  Can I ask you to standardize the format if you are going to insist on adding them?

mediator/med_phases_restart_mod.F90
```diff
@@ -52,7 +52,8 @@ subroutine med_phases_restart_write(gcomp, rc)
     type(ESMF_TimeInterval)    :: timediff       ! Used to calculate curr_time
     type(ESMF_Alarm)           :: alarm
     type(ESMF_Calendar)        :: calendar
-    character(len=64)          :: currtimestr, nexttimestr
+    character(len=64)          :: currtimestr
```
jedwards4b: Please don't hard code string lengths - use one of the already defined constants.

mediator/med_phases_restart_mod.F90
```diff
@@ -300,10 +304,21 @@ subroutine med_phases_restart_write(gcomp, rc)
                        nx=nx, ny=ny, nt=1, whead=whead, wdata=wdata, pre=trim(compname(n))//'ExpAccum', rc=rc)
                    if (ChkErr(rc,__LINE__,u_FILE_u)) return
                 endif
+
+                ! Write import field bundle accumulators
+                if (ESMF_FieldBundleIsCreated(is_local%wrap%FBImpAccum(n,n),rc=rc)) then
+                   ! TODO: only write this out if actually have done accumulation
```
jedwards4b: Should this comment be removed?

mvertens: yes - this is still needed. The accumulation field bundle is created by the present and active flags - but actual accumulation is triggered by run time calls.

mediator/med_phases_restart_mod.F90
mvertens: Done

mediator/med.F90
mvertens: I disagree. I think that it really helps readability. I agree about standardization - but its not clear to me what it should be. For now I've removed the extra decorations at mentioned in your comment.

mvertens:
@jedwards4b - I do not understand the statement you made above.
_Your change from med back to cpl is going to require a significant number of changes in cime._
The cime changes accompanying this PR only involve 5 files (see below) and except for the datm changes are very minor.  Could you please clarify.
```
config/cesm/config_grids.xml
scripts/lib/CIME/build.py
src/components/data_comps/datm/nuopc/atm_comp_nuopc.F90
src/components/data_comps/datm/nuopc/datm_comp_mod.F90
src/components/xcpl_comps/xglc/nuopc/glc_comp_nuopc.F90
src/share/timing/perf_mod.F90

mediator/med_phases_restart_mod.F90
mvertens: Done.

jedwards4b:
"med" is also found in system_tests_common.py, hist_utils.py, case_run.py, nodefail.py, ...

PR #50, opened 2020-01-29
remove stubs

jedwards4b:
Change Description: without stub models variables COMP_XXX are not defined - these changes allow for this condition. 

Testing Completed:
(Minimum testing: CIME_DRIVER=nuopc scripts_regression_tests.py)

CIME HASH:


mvertens:
@jedwards4b - I think this PR needs to be reworked. I think that we are at a place where we can revisit it. Do you agree?

jedwards4b:
dead code

Issue #74, opened 2020-07-02
use ESMF_Version_* to get version dependent differences in CMEPS


mvertens:
All of the ESMF_VERSION_* symbols are available as parameters via
"use ESMF" in Fortran code. This is documented under
https://esmf-org.github.io/dev_docs/ESMF_refdoc/node9.html#SECTION090161000000000000000.

Using the above Fortran named constants is the most portable way to code
version dependent differences in code because it does not require the
build system to get involved. I think this approach should be used
inside the CMEPS code for (ESMF_VERSION_MAJOR==8 .and.
ESMF_VERSION_MINOR>0), and it should replace the CPP #if based code.

There are cases where the CPP based #if approach is necessary -
basically for when API changes between ESMF versions will not allow
compilation under different versions. This is not the case here, and
therefore it will be much simpler to just use Fortran level logic based
on the named constants. The problem with the CPP macro based approach is
that it requires that the build system gets the variables from the
esmf.mk and then pass them in via -D option to the compiler. The NEMS
build system does this e.g. for module_EARTH_GRID_COMP.F90 and
module_MEDIATOR.F90, but not for CMEPS, which would need a change in the
CMEPS Makefile.

Again, my recommendation is to change the CMEPS version logic code to
using the ESMF_VERSION_* named constants defined by "use ESMF".

jedwards4b:
I think that there must be some confusion here.

jedwards4b:
The suggestion from Gerhard cannot be used since the ESMF_VERSION macro is being used to determine if elements of the ESMF module itself exist.   Therefore I think that this code needs to remain as is.
```
#if ESMF_VERSION_MINOR > 0                                                                           
    use ESMF , only : ESMF_StateSet, ESMF_StateIntent_Import, ESMF_StateIntent_Export                
    use ESMF , only : ESMF_StateIntent_Flag                                                          
#endif
```

PR #99, opened 2020-09-03
Fix CPP ESMF_VERSION_MAJOR and ESMF_VERSION_MINOR definitions.

danrosen25:
### Description of changes
Add CPPDEFS to mediator/Makefile
- ESMF_VERSION_MAJOR
- ESMF_VERSION_MINOR

Check ESMF_VERSION_MAJOR and ESMF_VERSION_MINOR
- mediator/esmFlds.F90
- mediator/med.F90

### Specific notes

none

use ESMF_Version_* to get version dependent differences in CMEPS #74

It's not a fix to this issue per se but version macros are not automatically passed to the preprocessor unless added to the CPPDEFS in mediator/Makefile. Also the code should check both MAJOR and MINOR versions.

 - [X] bit for bit

 - [X] No

- [ ] (required) CIME_DRIVER=nuopc scripts_regression_tests.py
   - machines:
- [ ] (required) CESM testlist_drv.xml
   - machines and compilers:
   - machines and compilers

Testing performed if application target is UFS-S2S:
- [ ] (required) UFS-S2S testing
   - description:

Hashes used for testing:
- [ ] CESM:
  - branch: nuopc_dev
  - hash:
- [ ] UFS-S2S, then umbrella repostiory to check out and associated hash:
  - repository to check out:
  - branch:
  - hash:


danrosen25:
@jedwards4b @uturuncoglu @mvertens 

Hi All,
I know you're thinking about replacing the CPP based ESMF VERSION checks but in the mean time please look at this fix. Question: How were you previously passing "-DESMF_VERSION_MAJOR=8 -DESMF_VERSION_MINOR=1" to the preprocessor?

Thanks.
Dan Rosen

jedwards4b:
These were provided in esmf.mk and used in the [cime Makefile](https://github.com/ESMCI/cime/blob/master/scripts/Tools/Makefile#L321)

jedwards4b (approved):

danrosen25:
Thanks @jedwards4b 
Just a note := will overwrite CPPDEFS, were you building with INTERNAL_PIO_INIT defined?

ifdef INTERNAL_PIO_INIT
CPPDEFS := -DINTERNAL_PIO_INIT
endif

jedwards4b:
@danrosen25 I've never used this Makefile and not sure who wrote it, but that looks like a mistake to me.

jedwards4b:
The cime based build doesn't use this Makefile.

rsdunlapiv:
We do need to make sure we preserve the `-DINTERNAL_PIO_INIT` since in UFS this is the only component using PIO, so the it needs to initialize PIO itself.  The line is question is the first definition of CCPDEFS, so shouldn't the `:=` be okay in that case?

danrosen25:
@rsdunlapiv 
Setting the CPPDEFS variable using `:=` is correct if CPPDEFS is only used locally but it will override global CPPDEFS. This PR will fix the standard GNUMake file, which isn't used in the CIME build but it is currently used by the HAFS build, and it will add a ESMF_VERSION_MAJOR check to the source code. So the PR won't fix everything but it's an improvement.

PR #122, opened 2020-10-29
performance upgrades - primarily creation of packed fields for regridding 

mvertens:
### Description of changes
performance upgrades - primarily creation of packed fields for regridding

### Specific notes
This PR does several things.
1. The main focus is to do all on the mapping on packed fields. For this to happen a new data type, 
```fortran
type, public :: packed_data_type
     integer, allocatable :: fldindex(:) ! size of number of packed fields
     character(len=CS)    :: mapnorm     ! normalization for packed field
     type(ESMF_Field)     :: field_src    ! packed sourced field
     type(ESMF_Field)     :: field_dst    ! packed destination field
     type(ESMF_Field)     :: field_fracsrc
     type(ESMF_Field)     :: field_fracdst
  end type packed_data_type
```
where fldindex:
- corresponds to the mapping of the packed field ungridded dimension to the index in the source field bundle for the target field
- has a size of the ungridded dimension of the packed field (which equals the number of fields that will be mapped for a given mapping type and source/destination pair) PLUS the number of fields with ungridded dimensions times the number of ungridded dimensions for each of those fields.
The new mapping occurs by copying all relevant fields from a source field bundle into a packed, mapping the packed field and then copying the relevant array data from the packed field back into the mapped field bundle. The cost of the copying is extremely small.

2. It turns out that querying for the rank of a field is costly and seemed to account for a lot of the cost of `med_merge_mod`. As a result, queries for the rank have in most cases been replaced for queries for the size of the ungridded dimension of a field, thereby significantly lowering the cost of the model run. Also -  `med_methods_mod.F90` has been reduced to remove unused methods as well as to replace the rank calls in most places.
3. Implemented the option to map atm->ocn and atm-ice uv fields with cart3d based on introducing a new mapping type mappatch_cart3d. This is turned on via the nuopc.config variable `mapuv_with_cart3d`  which by default is` .false.`

 - [X] bit for bit - all compses other than X 
 - [x] more substantial - only X compsets (changes to the glc fields)

 - [ ] Yes
 - [x] No

- [ ] (required) CIME_DRIVER=nuopc scripts_regression_tests.py
   - machines:
- [x] (required) CESM testlist_drv.xml
   - machines and compilers: cheyenne, intel
   - details (e.g. failed tests): No failed tests, compared to **oct26** baseline and generated **nov02** baselines in cmeps_baselines
   - machines and compilers

Testing performed if application target is UFS-S2S:
- [ ] (required) UFS-S2S testing
   - description:

Hashes used for testing:
- [x] CESM:
  - branch: nuopc_dev
  - hash: 2740d0c
- [ ] UFS-S2S, then umbrella repostiory to check out and associated hash:
  - repository to check out:
  - branch:
  - hash:


mediator/med_phases_prep_glc_mod.F90
```diff
@@ -892,36 +894,26 @@ subroutine med_phases_prep_glc_renormalize_smb(gcomp, rc)
...
-         srccomp=compglc, &
-         destcomp=complnd, &
-         FBSrc=FBglc_icemask, &
-         FBDst=FBlnd_icemask, &
-         FBFracSrc=FBglc_icemask, &  ! this will not be used since are asking for 'none' normalization
-         FBNormOne=is_local%wrap%FBNormOne(compglc,complnd,:), &
-         RouteHandles=is_local%wrap%RH(compglc,complnd,:), &
-         string='mapping Sg_imask_g to Sg_imask_l (from glc to land)', rc=rc)
+    ! map ice mask from glc to lnd with no normalization
+    ! BUG(wjs, 2017-05-11, #1516) I think we actually want norm = .false. here, but this needs more thought
```
jedwards4b, line 911: Is this comment still valid?  Don't see the referenced 'norm'

mediator/med_phases_prep_ocn_mod.F90
```diff
@@ -76,26 +76,34 @@ subroutine med_phases_prep_ocn_map(gcomp, rc)
...
 
     ! Count the number of fields outside of scalar data, if zero, then return
-    call FB_getNumFlds(is_local%wrap%FBExp(compocn), trim(subname)//"FBexp(compocn)", ncnt, rc)
-    if (ChkErr(rc,__LINE__,u_FILE_u)) return
+    call ESMF_FieldBundleGet(is_local%wrap%FBExp(compocn), fieldCount=ncnt, rc=rc)
+    if (chkerr(rc,__LINE__,u_FILE_u)) return
 
+    ! map all fields in FBImp that have active ocean coupling
     if (ncnt > 0) then
+       !DEBUG
```
jedwards4b, line 84: should this be removed?

mediator/med_phases_prep_ocn_mod.F90
```diff
@@ -76,26 +76,34 @@ subroutine med_phases_prep_ocn_map(gcomp, rc)
...
+                  field_normOne=is_local%wrap%field_normOne(n1,compocn,:), &
+                  packed_data=is_local%wrap%packed_data(n1,compocn,:), &
+                  routehandles=is_local%wrap%RH(n1,compocn,:), rc=rc)
              if (ChkErr(rc,__LINE__,u_FILE_u)) return
-          endif
-       enddo
+          end if
+       end do
+
+       !DEBUG
```
jedwards4b, line 102: Should this be removed?

jedwards4b (approved):

mediator/med_io_mod.F90
```diff
@@ -522,36 +526,27 @@ subroutine med_io_write_FB(filename, iam, FB, whead, wdata, nx, ny, nt, &
        return
     endif
 
-    call FB_getFieldN(FB, 1, field, rc=rc)
+    call ESMF_FieldGet(fieldlist(1), mesh=mesh, rc=rc)
```
uturuncoglu, line 529: This assumes that all the fields in the field list is in same mesh. Right?

mediator/med_io_mod.F90
```diff
@@ -601,72 +596,64 @@ subroutine med_io_write_FB(filename, iam, FB, whead, wdata, nx, ny, nt, &
        write(tmpstr,*) subname,' dimid = ',dimid
        call ESMF_LogWrite(trim(tmpstr), ESMF_LOGMSG_INFO)
 
+       ! TODO (mvertens, 2019-03-13): below is a temporary mod to NOT write hgt
        do k = 1,nf
-          call FB_getNameN(FB, k, itemc, rc=rc)
-          if (chkerr(rc,__LINE__,u_FILE_u)) return
+          if (trim(fieldnamelist(k)) == "hgt") then
```
uturuncoglu, line 601: Why this is here. If I remember correctly Sa_topo (or maybe I am wrong) was not matching when it is compared with the cprnc. Is this control for it?

mediator/med_map_mod.F90
```diff
@@ -594,529 +587,629 @@ logical function med_map_RH_is_created_RH1d(RHs,mapindex,rc)
...
-       ! -------------------
+             end if! end if source field is mapped to destination field with mapindex
+          end do ! end loop over FBSrc fields
+       end do ! end loop over fldsSrc elements
 
-       call ESMF_LogWrite(trim(subname)//" --> remapping "//trim(fldname)//" with "//trim(mapnames(mapindex)), &
-            ESMF_LOGMSG_INFO)
+       if (npacked(mapindex) > 0) then
+          ! Create the packed source field bundle for mapindex
+          allocate(ptrsrc_packed(npacked(mapindex), lsize_src))
```
uturuncoglu, line 856: It would be nice to check those are allocated or not. It is valid for all allocate statements.

mediator/med_map_mod.F90
```diff
@@ -594,529 +587,629 @@ logical function med_map_RH_is_created_RH1d(RHs,mapindex,rc)
...
+               meshloc=ESMF_MESHLOC_ELEMENT, rc=rc)
+          packed_data(mapindex)%field_fracdst = ESMF_FieldCreate(lmesh_dst, ESMF_TYPEKIND_R8, &
+               meshloc=ESMF_MESHLOC_ELEMENT, rc=rc)
+       end if
+    end do ! end loop over mapindex
 
-       if (mapindex == mapfcopy) then
-          call med_map_FB_Field_Regrid(FBSrc, fldname, FBDst, fldname, RouteHandles, mapindex, rc=rc)
-          if (chkerr(rc,__LINE__,u_FILE_u)) return
+    deallocate(npacked)
```
uturuncoglu, line 874: same for the deallocate statements. if those are already allocated then the deallocate can be called.

mediator/med_phases_ocnalb_mod.F90
```diff
@@ -102,6 +97,8 @@ subroutine med_phases_ocnalb_init(gcomp, ocnalb, rc)
     real(R8), pointer        :: ownedElemCoords(:)
```
uturuncoglu, line 97: since all pointers init to null(), it would be nice to init this one with null().

mediator/med_phases_prep_ice_mod.F90
```diff
@@ -53,16 +50,12 @@ subroutine med_phases_prep_ice(gcomp, rc)
     ! local variables
     type(ESMF_StateItem_Flag)      :: itemType
     type(InternalState)            :: is_local
+    type(ESMF_Field)               :: lfield
     integer                        :: i,n,n1,ncnt
     character(len=CS)              :: fldname
     integer                        :: fldnum
     integer                        :: mapindex
     real(R8), pointer              :: dataptr(:)
```
uturuncoglu, line 58: init with null()

uturuncoglu (commented):
@mvertens Let me test with the HAFS application before merge. I'll let you know about it. As we discussed with Jim yesterday, we might also want to update the PR template and include also testing with HAFS application. What do you think?

mediator/med_phases_prep_ice_mod.F90
uturuncoglu: There are also other cases that pointiest are not set to null() but I did not mark all of them.

DeniseWorthen:
Is this branch up to date w/ master? I don't see the recent changes in med_phases_aofluxes_mod.F90 where the 'first call' was moved into the init.

mediator/med_phases_prep_glc_mod.F90
mvertens: I've added a clarifying statement.

mediator/med_phases_prep_ocn_mod.F90
mvertens: Good catch. I've removed this.

mediator/med_phases_prep_ocn_mod.F90
mvertens: Removed as well. Thanks.

mediator/med_io_mod.F90
mvertens: Yes that is correct. All fields in a given fieldbundle are assumed to be on the same mesh. That is an underlying assumption in CMEPS.

mediator/med_map_mod.F90
mvertens: This allocate just occurs at initialization - so Its not clear to me why it would not be allocated. But I can add a check statement.

mvertens: But will this require rerunning all of the tests again?

mediator/med_map_mod.F90
mvertens: You need this deallocate for npacked since you are allocating and deallocating it as you go through the mapindices. I don't ever see a case where you would trigger a deallocate for an unallocated array. I'll double check.

mediator/med_phases_ocnalb_mod.F90
mvertens: Good point. I'm going to go through and set all pointers to null for consistency.

mediator/med_map_mod.F90
uturuncoglu: it is fine, we could do it later. I am testing it with HAFS now.

rsdunlapiv:
@DeniseWorthen would it be possible to have before and after timings if you test this?  The work from @mvertens should speed up mapping for UFS S2S as well.

DeniseWorthen:
> @DeniseWorthen would it be possible to have before and after timings if you test this? The work from @mvertens should speed up mapping for UFS S2S as well.

Sure; I'll take make a comparison.

mediator/med_io_mod.F90
mvertens: I don't think so. But frankly, I do not remember why it was put in. Maybe in the next PR or so I can address this question along with the other ones you raised. Can you please put in an issue for this.

mediator/med_map_mod.F90
mvertens: Okay - that would be great. Maybe you could raise an issue for this as well so that we don't forget to do it. Thanks for pointing this out.

mediator/med_phases_ocnalb_mod.F90
mvertens: I just pushed back all local pointers being set to null(). That is what Denise is testing with - and I will redo the tests tonight.

DeniseWorthen:
I have run 2 cases using the ufs-weather coupled control case at c384, 1/4 deg ocean and ice, comparing the perf update with the previous master. 

- The coupler restart files at 5d are b4b between these two runs

- Wall clock time in the two cases:
For the perf run: The total amount of wall time                        = 5748.449491
For the pre-perf run: The total amount of wall time                        = 5931.439371

Using this crude estimate, our 35d run might be ~20min faster.

uturuncoglu:
@mvertens Just let you know that the HAFS regression test (datm+hycom) passed.

mvertens:
@uturuncoglu @DeniseWorthen - when I reran the test suite last night I found unexpected issues with restart runs for some compsets. I think I found the problem and I'm rerunning the tests this morning.

rsdunlapiv:
@DeniseWorthen thanks for reporting on the timing.  Those are very encouraging numbers!  @mark-a-potts should be aware for this for his analysis of UFS S2S performance.

uturuncoglu (approved):

uturuncoglu:
@mvertens let me know if you need help for testing.

mvertens:
@uturuncoglu @DeniseWorthen - I have fixed all the problems and tested again on my end and all the tests pass now are are bfb with oct28. Could you please test again and make sure that nothing changes for you. Sorry to make you do this twice - but this is a really big PR so I want to make sure that everything works correctly.

DeniseWorthen:
It ran in nems_frac but didn't complete in nems_orig. I have a run here: /glade/scratch/worthen/hour2_escomp

DeniseWorthen (approved):

DeniseWorthen:
All tests in ufs are b4b w/ the previous master for this PR.

mvertens:
Please do not merge yet. My latest tests show unexpected answer changes for I2000. I also need to resolve the conflict in med_map_mod.F90 - not sure why this is there.

mvertens:
So it looks like the oct28 baseline is incorrect for the test ERS_Vnuopc_Ld5_P144.f19_f19_mg17.I2000Clm50Sp.cheyenne_intel.clm-default
The test from this PR is bfb with both the oct27 and the oct26 baselines in cmeps_baselines.
I will generate a new baseline and then I think the PR will be ready to merge.

PR #160, opened 2021-02-26
use pts_lat and pts_lon for single point cases

jedwards4b:
### Description of changes
Instead of defining a domain file with a single lat lon point we use the PTS_LAT and PTS_LON variables to input
the location to be used to create a mesh.   This requires the cime changes in https://github.com/ESMCI/cime/pull/3868

### Specific notes

 - [X] bit for bit
 - [ ] more substantial 

 - [ ] Yes
 - [ ] No

- [ ] (recommended) CIME_DRIVER=nuopc scripts_regression_tests.py
   - machines:
- [ ] (recommended) CESM testlist_drv.xml
   - machines and compilers:
   - machines and compilers
- [ ] (other) please described in detail
   - machines and compilers

Testing performed if application target is UFS-coupled:
- [ ] (recommended) UFS-coupled testing
   - description:

Testing performed if application target is UFS-HAFS:
- [ ] (recommended) UFS-HAFS testing
   - description:

Hashes used for testing:
- [ ] CESM:
  - branch: nuopc_dev
  - hash:
- [ ] UFS-coupled, then umbrella repostiory to check out and associated hash:
  - repository to check out:
  - branch:
  - hash:
- [ ] UFS-HAFS, then umbrella repostiory to check out and associated hash:
  - repository to check out:
  - branch:
  - hash:

### DIFF
--- a/cime_config/buildnml
+++ b/cime_config/buildnml
@@ -43,7 +43,8 @@ def _create_drv_namelists(case, infile, confdir, nmlgen, files):
     config['MPILIB'] = case.get_value('MPILIB')
     config['OS'] = case.get_value('OS')
     config['glc_nec'] = 0 if case.get_value('GLC_NEC') == 0 else case.get_value('GLC_NEC')
-    config['single_column'] = 'true' if case.get_value('PTS_MODE') else 'false'
+    if case.get_value("ATM_NX") == 1 and case.get_value("ATM_NY") == 1 and case.get_value("ATM_DOMAIN_FILE") != "UNSET":
+        config['single_column'] = 'true'
     config['timer_level'] = 'pos' if case.get_value('TIMER_LEVEL') >= 1 else 'neg'
     config['bfbflag'] = 'on' if case.get_value('BFBFLAG') else 'off'
     config['continue_run'] = '.true.' if case.get_value('CONTINUE_RUN') else '.false.'
--- a/cime_config/config_component.xml
+++ b/cime_config/config_component.xml
@@ -1233,21 +1233,13 @@
     <desc>grid mask - DO NOT EDIT (for experts only)</desc>
   </entry>
 
-  <entry id="PTS_MODE">
-    <type>logical</type>
-    <valid_values>TRUE,FALSE</valid_values>
-    <default_value>FALSE</default_value>
-    <group>run_domain</group>
-    <file>env_run.xml</file>
-    <desc>Operate on only a single point of the global grid  - DO NOT EDIT (for experts only)</desc>
-  </entry>
-
   <entry id="PTS_LAT">
     <type>real</type>
     <default_value>-999.99</default_value>
     <group>run_domain</group>
     <file>env_run.xml</file>
-    <desc>Latitude to find nearest points for points mode (only used if PTS_MODE is TRUE)</desc>
+    <desc>Latitude of grid location, in single column mode interpolate datasets to this location
+          in single point mode assume all datasets are at this location</desc>
   </entry>
 
   <entry id="PTS_LON">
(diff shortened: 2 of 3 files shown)


mvertens (changes requested):
Don't we need to set single_column accordingly as well here?
I thought the logic was that if single_column was was true we needed to use a domain file in order to do a nearest neighbor search - single_column_domainfile.
So how are we setting PTS_MODE to enforce that? This would be in config_component.xml.
Also - config_component.xml needs to be changed to rewrite the following comments for PTS_LON and PTS_LAT
`Longitude to find nearest points for points mode (only used if PTS_MODE is TRUE)`
` Latitude to find nearest points for points mode (only used if PTS_MODE is `TRUE)`
I am actually wondering if we want to get rid of points mode and JUST use the single_column_domainfile. If that is set then you will do a nearest neighbor search using PTS_LON and PTS_LAT otherwise you just use them. It might be easier than having both that filename AND the PTS_MODE to be set. Thoughts?

ekluzek:
> Don't we need to set single_column accordingly as well here?
> I thought the logic was that if single_column was was true we needed to use a domain file in order to do a nearest neighbor search - single_column_domainfile.
> So how are we setting PTS_MODE to enforce that? This would be in config_component.xml.
> Also - config_component.xml needs to be changed to rewrite the following comments for PTS_LON and PTS_LAT
> `Longitude to find nearest points for points mode (only used if PTS_MODE is TRUE)`
> `Latitude to find nearest points for points mode (only used if PTS_MODE is`TRUE)`
> I am actually wondering if we want to get rid of points mode and JUST use the single_column_domainfile. If that is set then you will do a nearest neighbor search using PTS_LON and PTS_LAT otherwise you just use them. It might be easier than having both that filename AND the PTS_MODE to be set. Thoughts?

Yes it seems to me we need to distinguish between running in single_column mode and running for a single point. single_column mode expects global datasets and then runs on the gridcell that's closest to the PTS_LAT,PTS_LON.

For single_column mode running with SCAM I think you DO want to use the global domain file because SCAM can bring in other information from the surrounding grid cells. For CTSM running on a single point the size, mask and area of the grid cell doesn't matter, but I'm not so sure of that for SCAM inside of CAM. They do have a mode of using files that bring in lateral advection from surrounding grid cells.

But, one thing that's critically important is that you you want single_column to be FALSE when running for a single point grid. You don't want to exercise the extra logic in finding the closest grid cells, especially because in CTSM that mode can NOT currently restart.

We probably need to bring in some SCAM people here.

cime_config/buildnml
```diff
@@ -43,7 +43,8 @@ def _create_drv_namelists(case, infile, confdir, nmlgen, files):
     config['MPILIB'] = case.get_value('MPILIB')
     config['OS'] = case.get_value('OS')
     config['glc_nec'] = 0 if case.get_value('GLC_NEC') == 0 else case.get_value('GLC_NEC')
-    config['single_column'] = 'true' if case.get_value('PTS_MODE') else 'false'
+    if case.get_value("ATM_NX") == 1 and case.get_value("ATM_NY") == 1 and case.get_value("ATM_DOMAIN_FILE") != "UNSET":
+        config['single_column'] = 'true'
```
ekluzek, line 47: We only want single_column = True for the case when you are running SCAM mode with global datasets. Turning single_column to true for cases of a single point exercises different code, that you don't need to. And it has limitations like CTSM can't restart. So we can't set it up this way. I really think that PTS_MODE needs to come back in. I like the idea of simplifying when we can, but I don't see that we can do that here.

jedwards4b:
@ekluzek I have tested both cases (scam and single point) successfully.  If you can produce a counter example I would be happy to consider it.

jedwards4b:
Superseded by #164

PR #174, opened 2021-03-30
add functionality for ocean precip scaling for CESM

mvertens:
### Description of changes
This change only effects CESM compsets

### Specific notes
This adds the capability currently only activated by CESM C and G compsets (with POP not MOM) to have a precip factor sent by POP to the mediator that is then applied to scale atmosphere rain and snow going back to the ocean and sea-ice (in this case POP and CICE/DICE). In the current testing, the precip factor was only set to 1 by POP - as a result, this PR does not change any baselines.

CMEPS Issues Fixed: None

 - [x] bit for bit
 - [ ] more substantial 

 - [ ] Yes
 - [x] No

- Verified manually that for SMS_Ld1_Vnuopc.T62_g17.G.cheyenne_intel.pop-default, the scaling factor to the mediator and that was set correctly in med_phases_prep_ocn_mod.F90 and that the results were bfb
- Verified that SMS_Vnuopc_Ld5.T62_t061.GMOM.cheyenne_intel was bfb before and after this change.

### DIFF
--- a/mediator/med_constants_mod.F90
+++ b/mediator/med_constants_mod.F90
@@ -11,6 +11,6 @@ module med_constants_mod
   real(R8), parameter :: med_constants_czero           = 0.0_R8  ! spval
   integer,  parameter :: med_constants_ispval_mask     = -987987 ! spval for RH mask values
   integer,  parameter :: med_constants_SecPerDay       = 86400   ! Seconds per day
-  integer             :: med_constants_dbug_flag       = 15
+  integer             :: med_constants_dbug_flag       = 0
 
 end module med_constants_mod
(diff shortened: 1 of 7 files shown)


DeniseWorthen (approved):
No impact to ufs-coupled or datm baselines.

uturuncoglu (approved):

mediator/med_constants_mod.F90
```diff
@@ -11,6 +11,6 @@ module med_constants_mod
   real(R8), parameter :: med_constants_czero           = 0.0_R8  ! spval
   integer,  parameter :: med_constants_ispval_mask     = -987987 ! spval for RH mask values
   integer,  parameter :: med_constants_SecPerDay       = 86400   ! Seconds per day
-  integer             :: med_constants_dbug_flag       = 0
+  integer             :: med_constants_dbug_flag       = 15
```
jedwards4b, line 14: change back to 0 here?

jedwards4b (changes requested):

jedwards4b (approved):

PR #185, opened 2021-05-07
Feature/multiple icesheet update

mvertens:
### Description of changes
Changes to actually send 2 ice sheets from CISM

### Specific notes
Verified that sending two identical ice sheets from CISM to the mediator had the mediator recognize them as separate but identical. This was with a TG compset.

CMEPS Issues Fixed: None

 - [x] bit for bit
 - [ ] more substantial

 - [ ] Yes
 - [x] No

aux_cism  test suite was run with nuopc with this change and was successful. 

Hashes used for testing:
- [x] CISM checkout using externals
  - repository to check out: https://github.com/escomp/cism-wrapper
  - tag:  cismwrap_2_1_81

### DIFF
--- a/cime_config/namelist_definition_drv.xml
+++ b/cime_config/namelist_definition_drv.xml
@@ -1694,6 +1694,18 @@
     </values>
   </entry>
 
+  <entry id="num_icesheets">
+    <type>integer</type>
+    <category>expdef</category>
+    <group>ALLCOMP_attributes</group>
+    <desc>
+      number of glc ice sheets
+    </desc>
+    <values>
+      <value>1</value>
+    </values>
+  </entry>
+
   <entry id="mapuv_with_cart3d">
     <type>logical</type>
     <category>mapping</category>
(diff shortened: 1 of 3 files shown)


DeniseWorthen (approved):
no foreseeable impact to ufs-model.

cime_config/namelist_definition_drv.xml
```diff
@@ -1694,6 +1694,18 @@
     </values>
   </entry>
 
+  <entry id="num_icesheets">
```
jedwards4b, line 1697: I'm wondering if this needs to be here - it seems like it may be better to query the cism model to tell you the number of ice sheets?

mediator/esmFlds.F90
```diff
@@ -27,11 +28,12 @@ module esmflds
          'ice ',&
          'rof ',&
          'wav ',&
-         'glc '/)
+         'glc1',&
```
jedwards4b, line 31: Is this a limitation that only allows up to 2 ice sheets?  If so it seems like a note to that effect should be added here.

jedwards4b (changes requested):
I think that the attribute for number of ice sheets should come from the cism namelist and config file not the driver one.

jedwards4b (approved):
https://github.com/ESCOMP/CMEPS/issues/186
satisfies my issues.  thanks

PR #189, opened 2021-05-09
Fix for budget annual budget output

### LINKED ISSUE #188: budgets are not written at the end of month 12 or at the end of year for a 1 year run
For a one year run, if the budget diagnostics are on and the annual and monthly budgets are requested, then no monthly budgets appear for December and no annual budgets appear.

mvertens:
### Description of changes
Fixes #188 where budgets are not written out for December or annually for a 1 year run

### Specific notes
The fix is to use nexttime rather than currtime to trigger budget output. This is consistent with the time stamp of history and restart files as well in cmeps.

Fixes: #188 

 - [x] bit for bit
 - [ ] more substantial

 - [ ] Yes
 - [x] No

This only effects CESM configurations. 
- Verified that an A compset at f19_g17 run for a year with budgets turned on produced both annual budget information as well as a budget entry for month 12.
- Verified that for a 3 month run, B1850 at f19_g17, budgets were created for month 3 and that the results were bfb with a run without these cmeps changes.

Hashes used for testing:
- [x] CESM:
  - tag: cesm2_3_beta02

### DIFF
--- a/mediator/med_diag_mod.F90
+++ b/mediator/med_diag_mod.F90
@@ -426,7 +426,7 @@ contains
    end subroutine med_diag_init
 
   !===============================================================================
-  subroutine med_diag_zero_mode(mode, rc )
+  subroutine med_diag_zero_mode(mode, rc)
 
     ! ------------------------------------------------------------------
     ! Zero out global budget diagnostic data.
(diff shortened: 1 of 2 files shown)


mediator/med_diag_mod.F90
```diff
@@ -421,95 +426,96 @@ end function get_diag_attribute
    end subroutine med_diag_init
 
   !===============================================================================
-  subroutine med_diag_zero( gcomp, mode, rc)
+  subroutine med_diag_zero_mode(mode, rc )
```
jedwards4b, line 429: remove the space after rc

jedwards4b (approved):

mediator/med_diag_mod.F90
mvertens: Done

DeniseWorthen (approved):
Should have no impact on ufs

Issue #231, opened 2021-09-22
restart problem in lnd->rof accum in cmeps0.13.27


mvertens:
lnd->rof accumulation fields are not being written to or read from the restarts

PR #232, opened 2021-09-24
fixed restart problem for aux_cam and aux_clm tests

mvertens:
### Description of changes
fix restart problem in lnd->rof accum in cmeps0.13.27 #231

### Specific notes
The lnd->rof accumulation fields are not being written to or read from the restarts with PR cmeps0.13.27 and this PR fixes that.
This will only effect CESM configurations.

CMEPS Issues Fixed: #231 

verified that cesm test ERS_Vnuopc_Ln9.f09_f09_mg17.F2000climo.cheyenne_intel.cam-outfrq9s

Hashes used for testing:
- [x] CESM:
  - tag: cesm2_3_beta05

### DIFF
--- a/mediator/med_phases_restart_mod.F90
+++ b/mediator/med_phases_restart_mod.F90
@@ -640,7 +640,7 @@ contains
     endif
 
     ! If lnd->glc, read accumulation from lnd to rof (CESM only)
-    if (ESMF_FieldBundleIsCreated(FBlndAccum2glc_r)) then
+    if (ESMF_FieldBundleIsCreated(FBlndAccum2rof_l)) then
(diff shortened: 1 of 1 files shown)


mvertens:
@uturuncoglu @DeniseWorthen - could you please have a quick look at this PR and let me know if we can merge it. @fischer-ncar - we need this update CMEPS tag to go into the next beta tag if we will point to the head of CMEPS (which I think would be good to do).

uturuncoglu (approved):
I did not tested with UFS but it seems that it won't affect at least HAFS.

DeniseWorthen (approved):
Nothing here should impact ufs.

PR #248, opened 2021-11-02
budget table updates needed for MOM6

mvertens:
### Description of changes
CESM budget table updates needed for MOM6

### Specific notes
This PR targets budget table fixes and updates needed when using MOM6 as the ocean component in CESM.
In particular a new salt budget table is used when MOM6 is present and is not there when POP is present.
This PR should not effect non-CESM applications since they currently do not use the budget module.

Contributors other than yourself, if any: @jedwards4b 

CMEPS Issues Fixed: None

Are changes expected to change answers? No - bit for bit.

Any User Interface Changes (namelist or namelist defaults changes)? Yes. Addition of `budget_table_version` to `nuopc.runconfig.` For POP, this is set to `v0` and for MOM6 its set to `v1`.

1. On cheyenne, ran SMS_Vnuopc_Ld5.f19_g17.B1850.cheyenne_intel.allactive-defaultio with and without the new changes and compared daily budget table outputs. 
2. On cheyenne, worked closely with with @gustavo-marques  to run various B compsets using MOM6 at a variety of resolutions to verify the correctness of the budgets.

Hashes used for testing:
- [x] CESM: cesm2_3_beta06

### DIFF
--- a/cime_config/namelist_definition_drv.xml
+++ b/cime_config/namelist_definition_drv.xml
@@ -919,6 +919,20 @@
     </values>
   </entry>
 
+  <entry id="budget_table_version">
+    <type>char</type>
+    <category>budget</category>
+    <group>MED_attributes</group>
+    <valid_values>v0,v1</valid_values>
+    <desc>
+      currently v0 refers to budgets using POP and v1 refers to budgets using MOM6
+    </desc>
+    <values>
+      <value comp_ocn="mom">v1</value>
+      <value comp_ocn="pop">v0</value>
+    </values>
+  </entry>
+
   <entry id="budget_inst">
     <type>integer</type>
     <category>budget</category>
@@ -2307,7 +2321,7 @@
     <group>ALLCOMP_attributes</group>
     <desc>
       If set to .true. BGC fields will be passed back and forth between the ocean and seaice
-      via the coupler.
+      via the mediator.
     </desc>
     <values>
       <value>.false.</value>
--- a/mediator/med_diag_mod.F90
+++ b/mediator/med_diag_mod.F90
@@ -24,13 +24,14 @@ module med_diag_mod
   use ESMF                  , only : ESMF_GridCompGet, ESMF_ClockGet, ESMF_TimeGet, ESMF_ClockGetNextTime
   use ESMF                  , only : ESMF_Alarm, ESMF_ClockGetAlarm, ESMF_AlarmIsRinging, ESMF_AlarmRingerOff
   use ESMF                  , only : ESMF_FieldBundle, ESMF_Field, ESMF_FieldGet
-  use shr_const_mod         , only : shr_const_rearth, shr_const_pi, shr_const_latice
+  use shr_const_mod         , only : shr_const_rearth, shr_const_pi, shr_const_latice, shr_const_latvap
   use shr_const_mod         , only : shr_const_ice_ref_sal, shr_const_ocn_ref_sal, shr_const_isspval
   use med_kind_mod          , only : CX=>SHR_KIND_CX, CS=>SHR_KIND_CS, CL=>SHR_KIND_CL, R8=>SHR_KIND_R8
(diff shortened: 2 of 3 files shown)


mediator/med_diag_mod.F90
```diff
@@ -698,9 +742,11 @@ subroutine med_phases_diag_atm(gcomp, rc)
...
 
     ! water isotopes
-    call diag_atm_wiso_send(is_local%wrap%FBImp(compatm,compatm), 'Faxa_evap_wiso', &
-         f_watr_evap_16O, f_watr_evap_18O, f_watr_evap_HDO, &
-         areas, lats, afrac, lfrac, ofrac, ifrac, budget_local, rc=rc)
+    if (flds_wiso) then
+       call diag_atm_wiso_send(is_local%wrap%FBImp(compatm,compatm), 'Faxa_evap_wiso', &
+            f_watr_evap_16O, f_watr_evap_18O, f_watr_evap_HDO, &
+            areas, lats, afrac, lfrac, ofrac, ifrac, budget_local, rc=rc)
+    end if
```
jedwards4b, line 749: there should be a if (ChkErr(rc,__LINE__,u_FILE_u)) return here.

mediator/med_diag_mod.F90
```diff
@@ -954,12 +1000,14 @@ subroutine med_phases_diag_lnd( gcomp, rc)
...
-    call diag_lnd_wiso(is_local%wrap%FBImp(complnd,complnd), 'Flrl_rofi_wiso', &
-         f_watr_ioff_16O, f_watr_ioff_18O, f_watr_ioff_HDO, ic, areas, lfrac, budget_local, rc=rc)
+    if (flds_wiso) then
+       call diag_lnd_wiso(is_local%wrap%FBImp(complnd,complnd), 'Flrl_evap_wiso', &
+            f_watr_evap_16O, f_watr_evap_18O, f_watr_evap_HDO, ic, areas, lfrac, budget_local, rc=rc)
+       call diag_lnd_wiso(is_local%wrap%FBImp(complnd,complnd), 'Flrl_rofl_wiso', &
+            f_watr_roff_16O, f_watr_roff_18O, f_watr_roff_HDO, ic, areas, lfrac, budget_local, rc=rc)
+       call diag_lnd_wiso(is_local%wrap%FBImp(complnd,complnd), 'Flrl_rofi_wiso', &
+            f_watr_ioff_16O, f_watr_ioff_18O, f_watr_ioff_HDO, ic, areas, lfrac, budget_local, rc=rc)
+    end if
```
jedwards4b, line 1010: each of these should have a if (ChkErr(rc,__LINE__,u_FILE_u)) return

jedwards4b (commented):
if (ChkErr(rc,__LINE__,u_FILE_u)) return  needs to be added throughout to handle rc

uturuncoglu (approved):

cime_config/namelist_definition_drv.xml
```diff
@@ -919,6 +919,20 @@
     </values>
   </entry>
 
+  <entry id="budget_table_version">
```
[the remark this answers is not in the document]
jedwards4b: right now its only for ocean differences but it could be for something else as well.   I think it should remain generic.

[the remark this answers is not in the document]
mvertens: @uturuncoglu - its not only for mom6 or pop at this point - its just a new version naming that I've added that can be used for any updates to the budget table. So for now v0 and v1 apply to pop and mom6 - but moving forwards there could be other changes that are not ocean specific that would need v2, etc.

mediator/med_diag_mod.F90
mvertens: @jedwards4b - thanks so much for catching this. There were a ton of places that I needed to add the error checks. I have done that and pushed things back.

jedwards4b (approved):

DeniseWorthen (approved):

PR #351, opened 2023-03-29
Cmeps sectorwater for pull request

TaranuDev:
### Description of changes
Here we add support for sectoral water abstraction module in CESM.
While the withdrawal, consumption and return flow fluxes for each of the sectors are computed in the land component (CLM/CTSM), the water is supplied from the routing model (MOSART in our case).

Therefore, there is a need to communicate these fluxes between the two models.

CMEPS Issues Fixed (include github issue #): 350

In the land component, there will be some small changes to the surface water-energy balance due to the fact that we apply the consumed water for sectoral usage on top of the soil columns covered by natural vegetation.

In the routing model, there will be some changes to the total water storage (VOLR) and final discharge into the oceans, since the sectoral water usage have a consumption element (meaning that the withdrawal is larger than the return flow, as some water is lost in the process).

Not in CMEPS specifically, but yes. 
In CTSM, we added some new namelist variables with default values already fixed. More about this in the dedicated pull request for the CTSM codebase.

### Testing performed
Extensive testing was performed, with different namelist configurations. Main goal was testing the model performance, then there are no inconsistencies between expected model behavior and results, but also model stability.
The model is behaving according to the design, and is completely stable (we performed simulations up to 30 years with no errors or issues).

For any questions, please contact me at sabin.taranu@vub.be.
I will be happy to assist to resolve any issues or anything unclear.

### DIFF
--- a/mediator/esmFldsExchange_cesm_mod.F90
+++ b/mediator/esmFldsExchange_cesm_mod.F90
@@ -3038,6 +3038,156 @@ contains
        end if
     end if
 
+        ! ---------------------------------------------------------------------
+    ! to rof: domestic withdrawal flux from land (withdrawal from rivers)
+    ! ---------------------------------------------------------------------
+    if (phase == 'advertise') then
+      call addfld(fldListFr(complnd)%flds, 'Flrl_dom_withd')
+      call addfld(fldListTo(comprof)%flds, 'Flrl_dom_withd')
+   else
+      if ( fldchk(is_local%wrap%FBImp(complnd, complnd), 'Flrl_dom_withd', rc=rc) .and. &
+           fldchk(is_local%wrap%FBExp(comprof)         , 'Flrl_dom_withd', rc=rc)) then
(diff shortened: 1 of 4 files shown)


ekluzek:
Because of the number of fields here, and because this won't be the default way of running, we will need a switch to activate passing these fields through the system. This should be a namelist item added to the drv_flds_in namelist file. That's the first thing I see from a first pass of looking through this.

wwieder:
Let's assess the scientific priority of the [CTSM side of this PR](https://github.com/ESCOMP/CTSM/pull/1975) before pushing forward on the externals associated with the work.

jedwards4b:
Please refer to the related [CTSM PR ](https://github.com/ESCOMP/CTSM/pull/1975) for further discussion and timeline.  I am converting this to draft status for now.

samsrabin:
Wanted to note here that I'm the assignee for the parent PR, ESCOMP/CTSM#1975.

PR #372, opened 2023-05-03
[WIP] add nan checks to prep modules

### LINKED ISSUE #370: Add NaN checks for coupling fields
In the MCT-based CPL7, we had a subroutine that checked the fields sent between components and the coupler to detect NaN values and abort if any were found:

https://github.com/ESCOMP/CESM_CPL7andDataComps/blob/d16b2de2e759ab52b94cf0b17fa4f788931ecd4f/driver/main/component_type_mod.F90#L229-L263

@cacraigucar pointed out that we don't have a similar check in CMEPS, but would like to have one.

I talked a few weeks ago with members of the ESMF team. The feeling at that meeting was that, while long-term it could make sense to implement this inside ESMF itself, that would take a fair amount of work because of the need to do it in a general way that works for all systems. So the best near-term approach is probably to implement something like the above subroutine in CMEPS. A piece that could make this more challenging than in CPL7 is determining where this check should go - i.e., where we can intercept fields at the right point to do this check.

An alternative – less ideal but a possible fall-back if there isn't a good interception point in CMEPS – would be for components to add these checks in their caps. CTSM already does this, using this subroutine:

https://github.com/ESCOMP/CTSM/blob/0b98af47bcd307ca04ceaa0ed03c254d6be0344e/src/cpl/utils/lnd_import_export_utils.F90#L143-L167

so we could move that subroutine into a shared location so that other components could also call it from their caps. (The downsides of that approach are that it would require changes in all components that want it and it could lead to less consistency across the system in terms of when this is called.)

jedwards4b:
### Description of changes
Check for Nans in med_phases_prep modules

### Specific notes

Resolves #370 

### Testing performed

Testing performed if application target is CESM:
- [ ] (recommended) CIME_DRIVER=nuopc scripts_regression_tests.py
   - machines:
- [ ] (recommended) CESM testlist_drv.xml
   - machines and compilers:
   - machines and compilers
- [ ] (other) please described in detail
   - machines and compilers

Testing performed if application target is UFS-coupled:
- [ ] (recommended) UFS-coupled testing
   - description:

Testing performed if application target is UFS-HAFS:
- [ ] (recommended) UFS-HAFS testing
   - description:

### Hashes used for testing:

- [ ] CESM:
  - branch/hash:
- [ ] UFS-coupled, then umbrella repostiory to check out and associated hash:
  - repository to check out:
  - branch/hash:
- [ ] UFS-HAFS, then umbrella repostiory to check out and associated hash:
  - repository to check out:
  - branch/hash:

### DIFF
--- a/mediator/med_methods_mod.F90
+++ b/mediator/med_methods_mod.F90
@@ -24,6 +24,11 @@ module med_methods_mod
     med_methods_FieldPtr_compare2
   end interface
 
+  interface med_methods_check_for_nans
+     module procedure med_methods_check_for_nans_1d
+     module procedure med_methods_check_for_nans_2d
+  end interface med_methods_check_for_nans
+
   ! used/reused in module
 
   logical                       :: isPresent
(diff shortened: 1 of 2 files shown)


DeniseWorthen:
Would this additional check be expected to have much performance impact?

jedwards4b:
Thanks for asking @DeniseWorthen : We will confirm the performance impact and if it's significant we will either only turn on in debug mode or provide a variable for the user to toggle on or off.

klindsay28:
It doesn't look like you're setting `rc` when no NaNs are found.

Also, I think it would be helpful in `med_phases_prep_atm` to check all fields for NaNs and then returning if NaN is found in any field, instead of returning after the first detection of NaN. That way, when more than one field has NaNs, you get a message for all of them. I think can be helpful when debugging such an occurrence.

mvertens (changes requested):
This looks good. Implementing the addition that Keith suggested would be good. However, we also want an interface where we pass in a field bundle - and it goes through each field, determines if the field has an undistributed dimension and then extracts the 1d or 2d pointer out and then calls this check routine.

billsacks:
Thank you very much for your work on this feature, @jedwards4b ! I have edited your top-level description to note that this will resolve #370.

jedwards4b:
Superseded by #377

PR #377, opened 2023-05-05
checking of nans in export field bundles

mvertens:
### Description of changes
checking of nans in export field bundles #377

### Specific notes
Introduced a new method med_methods_FB_check_for_nans in med_methods_mod and has all of the med_phases_prep_xxx_mod.F90 files call this method. If Nans are found in any field the model will abort, but it will first print out the number of nans found for each field and on each processor.

Contributors other than yourself, if any: @jedwards4b 

CMEPS Issues Fixed: #370 

Are changes expected to change answers?  bfb

### Testing performed
Verified that if nans were set in the 1d and 2d field in the call to med_phases_prep_atm, then PET files were written with the right statistics and the model aborted.

### DIFF
--- a/mediator/med_phases_prep_glc_mod.F90
+++ b/mediator/med_phases_prep_glc_mod.F90
@@ -34,6 +34,7 @@ module med_phases_prep_glc_mod
   use med_methods_mod       , only : fldbun_diagnose  => med_methods_FB_diagnose
   use med_methods_mod       , only : fldbun_reset     => med_methods_FB_reset
   use med_methods_mod       , only : fldbun_init      => med_methods_FB_init
+  use med_methods_mod       , only : fldbun_check_for_nans => med_methods_FB_check_for_nans
   use med_methods_mod       , only : field_getdata2d  => med_methods_Field_getdata2d
   use med_methods_mod       , only : field_getdata1d  => med_methods_Field_getdata1d
   use med_utils_mod         , only : chkerr           => med_utils_ChkErr
@@ -705,6 +706,12 @@ contains
           end do
        endif
     end if
+
+    ! Check for nans in fields export to atm
+    do ns = 1,is_local%wrap%num_icesheets
+       call fldbun_check_for_nans(is_local%wrap%FBExp(compglc(ns)), rc=rc)
+       if (ChkErr(rc,__LINE__,u_FILE_u)) return
+    end do
 
     if (dbug_flag > 5) then
        call ESMF_LogWrite(trim(subname)//": done", ESMF_LOGMSG_INFO)
--- a/mediator/med_phases_prep_glc_mod.F90
+++ b/mediator/med_phases_prep_glc_mod.F90
@@ -34,7 +34,7 @@ module med_phases_prep_glc_mod
   use med_methods_mod       , only : fldbun_diagnose  => med_methods_FB_diagnose
   use med_methods_mod       , only : fldbun_reset     => med_methods_FB_reset
   use med_methods_mod       , only : fldbun_init      => med_methods_FB_init
-  use med_methods_mod       , only : fldbun_check_for_nans => med_methods_FB_check_for_nans
+  use med_methods_mod       , only : FB_check_for_nans => med_methods_FB_check_for_nans
   use med_methods_mod       , only : field_getdata2d  => med_methods_Field_getdata2d
   use med_methods_mod       , only : field_getdata1d  => med_methods_Field_getdata1d
   use med_utils_mod         , only : chkerr           => med_utils_ChkErr
@@ -709,7 +709,7 @@ contains
 
     ! Check for nans in fields export to atm
     do ns = 1,is_local%wrap%num_icesheets
-       call fldbun_check_for_nans(is_local%wrap%FBExp(compglc(ns)), rc=rc)
+       call FB_check_for_nans(is_local%wrap%FBExp(compglc(ns)), rc=rc)
        if (ChkErr(rc,__LINE__,u_FILE_u)) return
     end do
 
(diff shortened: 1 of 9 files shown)


mediator/med_phases_prep_lnd_mod.F90
```diff
@@ -29,6 +29,7 @@ subroutine med_phases_prep_lnd(gcomp, rc)
     use ESMF                  , only : ESMF_StateGet, ESMF_StateItem_Flag, ESMF_STATEITEM_NOTFOUND
     use esmFlds               , only : med_fldList_GetFldListTo, med_fldList_type
     use med_methods_mod       , only : fldbun_diagnose  => med_methods_FB_diagnose
+    use med_methods_mod       , only : fldbun_check_for_nans => med_methods_FB_check_for_nans
```
jedwards4b, line 32: FB_check_for_nans?

mediator/med_phases_prep_rof_mod.F90
```diff
@@ -23,6 +23,7 @@ module med_phases_prep_rof_mod
   use med_methods_mod       , only : fldbun_reset     => med_methods_FB_reset
   use med_methods_mod       , only : fldbun_average   => med_methods_FB_average
   use med_methods_mod       , only : field_getdata1d  => med_methods_Field_getdata1d
+  use med_methods_mod       , only : fldbun_check_for_nans => med_methods_FB_check_for_nans
```
jedwards4b, line 26: FB_check_for_nans

mediator/med_phases_prep_glc_mod.F90
```diff
@@ -34,6 +34,7 @@ module med_phases_prep_glc_mod
   use med_methods_mod       , only : fldbun_diagnose  => med_methods_FB_diagnose
   use med_methods_mod       , only : fldbun_reset     => med_methods_FB_reset
   use med_methods_mod       , only : fldbun_init      => med_methods_FB_init
+  use med_methods_mod       , only : fldbun_check_for_nans => med_methods_FB_check_for_nans
```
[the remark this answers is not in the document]
mvertens: DONE

DeniseWorthen:
I will test UFS, but I'd like to wait until after the WW3 unstructured mesh gets committed early this week.

mvertens:
@DeniseWorthen - thanks! 

@uturuncoglu @denise - This should have no impact on the UFS since I have inserted an #ifdef for this to only be relevant in CESM. The issue is that there is no code in UFS that implements the shr_infnan_mod logic that we are using in CESM. So until that type of functionality exists - UFS should just call this routine and immediately return.

@uturuncoglu - for CMCC this code can be activated since they are also using the share code that CESM is using.

DeniseWorthen:
@mvertens Thanks. I hadn't actually looked at the changes yet ;-)

jedwards4b (approved):

billsacks:
Thanks a lot for your work on this, @mvertens - this will be very nice to have!

DeniseWorthen (approved):
@mvertens We have the shr_infnan_mod.F90 available in our CDEPS repo so I will try to get this working on our side for a later PR.

DeniseWorthen:
@jedwards4b In med_io_mod, we have
```
use shr_infnan_mod, only: shr_infan_isnan
```
should that be 
```
use shr_infnan_mod, only: shr_infnan_isnan
```

?

DeniseWorthen:
@mvertens @jedwards4b I've been able to turn this feature on for UFS now but I think I'm seeing a fairly big impact on run time. I ran our c96-1deg ocean+ice case for 5 days with this feature on and off. To turn off, I just inserted a return at the top of the ``med_methods_FB_check_for_nans``. The wall clock times were: nan check on 1220s, nan check off 986s. 

Was a similar impact seen in CESM?

I'm wondering how to best make this optional for UFS, since it would be a nice feature to have available. One way I thought is 

```
diff --git a/mediator/med_methods_mod.F90 b/mediator/med_methods_mod.F90
index 1da8d6ac..4425e606 100644
--- a/mediator/med_methods_mod.F90
+++ b/mediator/med_methods_mod.F90
@@ -2530,6 +2530,11 @@ contains
     ! ----------------------------------------------
     rc = ESMF_SUCCESS

+#ifndef CESMCOUPLED
+    if (dbug_flag == 0) then
+       return
+    end if
+#endif
     call ESMF_FieldBundleGet(FB, fieldCount=fieldCount, rc=rc)
```

Or, there could be a config variable to toggle. Or, it could be controlled w/ a #ifdef DEBUG. Any thoughts?

jedwards4b:
@DeniseWorthen Thanks for looking at the performance impact of this change.  I will run a similar test with PFS.ne30pg3_t061.B1850MOM and let you know.  I think it would be best to add a namelist variable to control it - I'll open a issue to track that.

jedwards4b:
For a 20 day run with PFS.ne30pg3_t061.B1850MOM
with nan checks:       3.36 ypd
without nan checks:  3.51 ypd

mvertens:
@jedwards4b - thanks for this test. I think its worth keeping this check with this small of a hit- particularly since we don't turn on floating point trapping in production runs. Do you agree? I think it also should be discussed at a CSEG meeting to make sure there is agreement to keep this given that there is a cost.

DeniseWorthen:
Another follow-up. It turns out that the impact UFS was seeing in run time w/ this feature on was *not* due to the NaN checking, but w/ the switching of the default PIO settings in an earlier PR (#378). 

If PIO settings are not provided, med_io had been using 

```
       cvalue = 'BOX'
       pio_rearranger = PIO_REARR_BOX
```

This was changed to 

```
       cvalue = 'SUBSET'
       pio_rearranger = PIO_REARR_SUBSET
```

If I revert that change, turning on the check-for-nan feature has negligible impact on wall-clock for our 5day test (in fact, it was a tiny bit faster, but that is just noise). 

I don't understand the PIO settings and how to choose them, but it looks I can specify our old settings as a configuration using

``pio_rearranger = SUBSET``

jedwards4b:
@DeniseWorthen Thanks for looking into this further.   The subset rearranger scales better but is slower at small task counts.

DeniseWorthen:
@jedwards4b Interesting! Gerhard has been working w/ Jun on various scaling issues w/ the coupled model and one of the issues he highlighted was both CMEPS and CICE slow I/O. We have not been using PIO for CICE but started to look into it after Gerhard's evaluations. It sounds like we should test using the subset rearranger also.

Issue #384, opened 2023-05-18
Add a namelist variable to control check_for_nans feature 


jedwards4b:
The new [check for nans ](https://github.com/ESCOMP/CMEPS/pull/377) feature may have a significant performance impact.   We should add a namelist control variable to turn this feature on and off.

PR #386, opened 2023-05-19
Add check for nans config option

jedwards4b:
### Description of changes
Add check_for_nans to the nuopc.runconfig 

### Specific notes

CMEPS Issues Fixed: #384

Are changes expected to change answers? (specify if bfb, different at roundoff, more substantial)  BFB

Any User Interface Changes (namelist or namelist defaults changes)?  Add
logical variable check_for_nans modifiable in user_nl_cpl.

### Testing performed
Tested using cesm2_3_alpha13b
PFS.ne30pg3_t061.B1850MOM.cheyenne_intel.allactive-defaultio

### DIFF
--- a/mediator/med.F90
+++ b/mediator/med.F90
@@ -661,6 +661,7 @@ contains
     use NUOPC , only : NUOPC_CompAttributeGet, NUOPC_CompAttributeSet, NUOPC_CompAttributeAdd
     use esmFlds, only : med_fldlist_init1, med_fld_GetFldInfo, med_fldList_entry_type
     use med_phases_history_mod, only : med_phases_history_init
+    use med_methods_mod       , only : mediator_checkfornans  
 
     ! input/output variables
     type(ESMF_GridComp)  :: gcomp
@@ -915,6 +916,20 @@ contains
           end do
        end if
     end do ! end of ncomps loop
+
+    ! Should mediator check for NaNs?
+    call NUOPC_CompAttributeGet(gcomp, name="check_for_nans", value=cvalue, rc=rc)
+    if (ChkErr(rc,__LINE__,u_FILE_u)) return
+    read(cvalue, *) mediator_checkfornans       
+    if(maintask) then
+       write(logunit,*) ' check_for_nans is ',mediator_checkfornans
+       if(mediator_checkfornans) then
+          write(logunit,*) ' Fields will be checked for NaN values when passed from mediator to component'
+       else
+          write(logunit,*) ' Fields will NOT be checked for NaN values when passed from mediator to component'
+       endif
+    endif
+
 
     if (profile_memory) call ESMF_VMLogMemInfo("Leaving "//trim(subname))
     call ESMF_LogWrite(trim(subname)//": done", ESMF_LOGMSG_INFO)
--- a/mediator/med.F90
+++ b/mediator/med.F90
@@ -918,9 +918,13 @@ contains
     end do ! end of ncomps loop
 
     ! Should mediator check for NaNs?
-    call NUOPC_CompAttributeGet(gcomp, name="check_for_nans", value=cvalue, rc=rc)
+    call NUOPC_CompAttributeGet(gcomp, name="check_for_nans", value=cvalue, isPresent=isPresent, isSet=isSet, rc=rc)
     if (ChkErr(rc,__LINE__,u_FILE_u)) return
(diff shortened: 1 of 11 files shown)


jedwards4b:
I've introduced a circular dependency - will need to address.

mediator/med_methods_mod.F90
```diff
@@ -2506,12 +2506,14 @@ subroutine med_methods_FB_getmesh(FB, mesh, rc)
   end subroutine med_methods_FB_getmesh
 
   !-----------------------------------------------------------------------------
-  subroutine med_methods_FB_check_for_nans(FB, rc)
-
-    use ESMF, only : ESMF_FieldBundle, ESMF_Field, ESMF_FieldBundleGet, ESMF_FieldGet
-
+  subroutine med_methods_FB_check_for_nans(gcomp, FB, maintask, logunit, rc)
```
mvertens, line 2509: I don't understand why you don't set checkfornans as a module variable - and introduce another module variable that is determine_checkfornans. If determine_checkfornans is .false. - then set checkfornans. I would think that would be cleaner than adding a save attribute to the logicals (firstcall and checkfornans).

mediator/med_phases_prep_atm_mod.F90
```diff
@@ -245,7 +245,7 @@ subroutine med_phases_prep_atm(gcomp, rc)
     end if
 
     ! Check for nans in fields export to atm
-    call FB_check_for_nans(is_local%wrap%FBExp(compatm), rc=rc)
+    call FB_check_for_nans(gcomp, is_local%wrap%FBExp(compatm), maintask, logunit, rc=rc)
```
mvertens, line 248: I don't understand why you are passing gcomp here. It seems that you need an initialization call to determine checkfornans as a module variable in med_methods_mod and then you don't need to pass any gcomp to the subroutines.
At least that is what I am subbesting.

billsacks:
Thanks for adding this flag @jedwards4b ! Given the performance impact documented in https://github.com/ESCOMP/CMEPS/pull/377#issuecomment-1553595338 (4.5% slowdown for a typical production case), I would vote for making the default be to *not* check for NaNs - or maybe to make the default be to check for NaNs in debug cases but not in production cases. My thinking is that, in most cases, NaNs will be detected eventually, so this CMEPS check is more about detecting them as soon as possible... so people can always rerun with the check enabled in these situations.

mediator/med.F90
```diff
@@ -916,6 +917,20 @@ subroutine AdvertiseFields(gcomp, importState, exportState, clock, rc)
        end if
     end do ! end of ncomps loop
 
+    ! Should mediator check for NaNs?
```
DeniseWorthen, line 920: should there be a isPresent, isSet check here?

jedwards4b: Sure I can add that and default to false if not present and set.  Thanks

mediator/med_methods_mod.F90
```diff
@@ -2532,9 +2532,10 @@ subroutine med_methods_FB_check_for_nans(FB, rc)
 
 #ifndef CESMCOUPLED
     ! For now only CESM uses shr_infnan_isnan - so until other models provide this
-    RETURN
+    mediator_checkfornans = .false.
 #endif
-
+    if(.not. mediator_checkfornans) return
+
```
DeniseWorthen, line 2538: With the change in how you set the config variable, we can remove the CESMCOUPLED ifdef here.

jedwards4b: The reason I haven't done that yet is that the github ext test doesn't include the cdeps library and thus the share code.  I am working on updating the github test to include that and then will remove the ifdefs once I have that working.  I got pulled into another issue though so it may be a day or two before I can get back to this, for now I'll convert this PR to draft.

DeniseWorthen: OK, thanks. I was worried I was holding you up by not testing this yet.

jedwards4b:
I have added the cdeps build to the github workflow and removed the CESMCOUPLED ifdefs from med_methods_mod.F90.   I think this is ready to merge.

DeniseWorthen (approved):
This is working for UFS with the flag to check either on or off. Thanks.

mvertens (approved):
Looks good. Thanks for the new implementation.

PR #404, opened 2023-08-29
add med_enthaply_mod

jedwards4b:
### Description of changes
Update calculation of enthalpy and include for F cases.
### Specific notes

Are changes expected to change answers? yes

Added compute_enthalpy to driver namelist. 

### Testing performed

### DIFF
--- a/mediator/med_enthalpy_mod.F90
+++ b/mediator/med_enthalpy_mod.F90
@@ -1,0 +1,171 @@
+module med_enthalpy_mod
+  use ESMF, only : ESMF_SUCCESS, ESMF_GridComp, ESMF_VMAllreduce, ESMF_REDUCE_SUM
+  use shr_kind_mod, only : R8=>shr_kind_r8
+  use shr_const_mod, only : tkfrz=>shr_const_tkfrz, cpfw=>shr_const_cpfw, cpice=>shr_const_cpice,&
+       cpwv=>shr_const_cpwv, cpsw=>shr_const_cpsw, pi=>shr_const_pi
+  use med_utils_mod         , only : chkerr        => med_utils_ChkErr
+  use med_methods_mod       , only : FB_fldchk     => med_methods_FB_FldChk
+  use med_methods_mod       , only : FB_GetFldPtr  => med_methods_FB_GetFldPtr
+  use med_internalstate_mod, only : compocn, compatm, InternalState
+  use perf_mod, only : t_startf, t_stopf
+    
+
+  implicit none
+  public :: med_compute_enthalpy
+
+  real(r8) :: global_htot_corr(1)
+  character(*), parameter :: u_FILE_u  = &
+       __FILE__
+contains
+  real(r8) function med_enthalpy_get_global_htot_corr()
+    ! Just return the latest global_htot_corr
+    med_enthalpy_get_global_htot_corr = global_htot_corr(1)
+  end function med_enthalpy_get_global_htot_corr
+
+  subroutine med_compute_enthalpy(is_local, rc)
+    type(InternalState), intent(in) :: is_local
+    integer, intent(out) :: rc
+
+    ! local variables
+    
+    real(r8), pointer :: tocn(:), rain(:), snow(:), rofl(:), rofi(:), evap(:)
+    real(r8), pointer :: rainl(:), rainc(:)
+    real(r8), pointer :: snowl(:), snowc(:)
+    real(r8), pointer :: hrain(:), hsnow(:), hevap(:), hcond(:), hrofl(:), hrofi(:)
+    real(r8), allocatable :: hcorr(:)
+    real(r8), pointer :: areas(:)
+    real(r8), parameter :: glob_area_inv = 1._r8 / (4._r8 * pi)
+    real(r8) :: local_htot_corr(1)
+    integer :: n, nmax
+    character(len=*), parameter:: subname = "med_compute_enthalpy"
+
+    call t_startf(subname)
+    rc = ESMF_SUCCESS
+    nmax = size(tocn)
+
+    call FB_GetFldPtr(is_local%wrap%FBImp(compocn,compocn), 'So_t', tocn, rc=rc)
+    if (ChkErr(rc,__LINE__,u_FILE_u)) return
+
+    if(FB_fldchk(is_local%wrap%FBExp(compocn), 'Faxa_rain', rc)) then
+       call FB_GetFldPtr(is_local%wrap%FBExp(compocn), 'Faxa_rain' , rain, rc=rc)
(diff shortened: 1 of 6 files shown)


jedwards4b:
Tested against b.e23_alpha16b.BLT1850.ne30_t232.040. results are bfb.

jedwards4b:
This is now functional for an f case.  I added a print of the global enthalphy correction value to the mediator log file.  Values over a 5 day run of ne30pg3_ne30pg3_mg17.FMTHIST_v0d are:
 0.0 on the first step
-0.66 on the second step
then ranging between 0 and 1 over the rest of the run.

jedwards4b:
Values for the B case are very similar.

mediator/med_enthalpy_mod.F90
```diff
@@ -0,0 +1,175 @@
...
+    else
+       allocate(hrofi(nmax))
+    endif
+    do n = 1,nmax
+       ! Need max to ensure that will not have an enthalpy contribution if the water is below 0C
+       hrain(n)  = max((tocn(n) - tkfrz), 0._r8) * rain(n)  * cpfw
+       hsnow(n)  = min((tocn(n) - tkfrz), 0._r8) * snow(n)  * cpice
+       hevap(n)  = (tocn(n) - tkfrz) * min(evap(n), 0._r8)  * cpwv
+       hcond(n)  = (tocn(n) - tkfrz) * max(evap(n), 0._r8)  * cpwv
+       hrofl(n)  = max((tocn(n) - tkfrz), 0._r8) * rofl(n)  * cpsw
```
gustavo-marques, line 143: Please change
`hrofl(n)  = max((tocn(n) - tkfrz), 0._r8) * rofl(n)  * cpsw
`
to 
`hrofl(n)  = max((tocn(n) - tkfrz), 0._r8) * rofl(n)  * cpfw`

mediator/med_enthalpy_mod.F90
```diff
@@ -0,0 +1,175 @@
...
+    if(FB_fldchk(is_local%wrap%FBExp(compocn), 'Foxx_hrofi', rc)) then
+       call FB_GetFldPtr(is_local%wrap%FBExp(compocn), 'Foxx_hrofi', hrofi, rc=rc)
+       if (ChkErr(rc,__LINE__,u_FILE_u)) return
+    else
+       allocate(hrofi(nmax))
+    endif
+    do n = 1,nmax
+       ! Need max to ensure that will not have an enthalpy contribution if the water is below 0C
+       hrain(n)  = max((tocn(n) - tkfrz), 0._r8) * rain(n)  * cpfw
+       hsnow(n)  = min((tocn(n) - tkfrz), 0._r8) * snow(n)  * cpice
```
gustavo-marques, line 140: Please change
```
hrain(n)  = max((tocn(n) - tkfrz), 0._r8) * rain(n)  * cpfw
￼hsnow(n)  = min((tocn(n) - tkfrz), 0._r8) * snow(n)  * cpice
```
to 
```
hrain(n)  = max((tbot(n) - tkfrz), 0._r8) * rain(n)  * cpfw
￼hsnow(n)  = min((tbot(n) - tkfrz), 0._r8) * snow(n)  * cpice
```
where **tbot** is 

```
if  (FB_fldchk(is_local%wrap%FBExp(compocn), 'Sa_tbot'    , rc=rc)) then
         call FB_GetFldPtr(is_local%wrap%FBExp(compocn), 'Sa_tbot', tbot, rc=rc)
endif
```

gustavo-marques (changes requested):

mediator/med_enthalpy_mod.F90
jedwards4b: done

mediator/med_enthalpy_mod.F90
jedwards4b: done - Note that with these changes it is no longer possible to compare to b.e23_alpha16b.BLT1850.ne30_t232.040

mvertens (approved):
The nems changes need to be reverted - otherwise this is fine!

jedwards4b:
I'm not sure who will do the comparison - I don't feel qualified.   But I think that you want to compare to the
F case you are used to looking at and so the tuning should be included.

tto061:
I think there are several problems with this code.

Gustavo and Peter and I have been discussing (a lot) on the scientifically correct and technically best way to implement the exchange material enthalpy fluxes. The main points we agreed on are:

1. hrain and hsnow depend on atmospheric processes and can only be computed correctly in the atmospheric model component; for consistency also hevap should be computed there

2. is it in fact not even possible to maintain energy consistency and (a fortiori) conservation in the atmosphere unless this is done

3. so the mediator is the wrong place for these calculations; in particular the current calculation breaks atmospheric energy conservation even at a basic level because it uses a zero meterial enthalpy point of 0°C: the fluxes passed to each component always need to be re-referenced to the zero enthalpy point of that component

4. even before all this, we had already decided to update the mediator to use the correct heat capacities for each water species

5. finally, with atmospheric energy consistency in place, we only need to use the mediator fixer (g.a. correction to the sensible heat flux) for the run-off components (hrofi/l)


I have already written code, tested it (both on betzy and on cheyenne) and shared it with Gustavo and Peter that does what we regard to be the correct calculation, computing the hfluxes in the atmosphere model (although not in the final way we eventually wish it to be done) and passing them to the ocean. My code works in all F, G, and coupled cases, (only for G cases the hfluxes are computed in the way you do).

As Gustavo and Peter already know, you can find this as SourceMods in my case directory
/glade/work/toniazzo/4gustavo/cases/G/g.e23_b15.GJRAv4.TL319_t232_zstar_N65.toniazzo
(for now it does still lack the namelist and diagnostic output parts). I invited G&P to check and run tests of their own (I don't know if they did).

So it loosk to me like there's been some poor communication, and ill timing, here. I think this pull request should be revised. I would suggest to try a merge with what I've written, and I am entirely at your disposal for questions and to help out.

gustavo-marques (approved):

jedwards4b:
@tto061 please review: https://github.com/jedwards4b/CMEPS/pull/2

gustavo-marques:
I am trying to run a G compset and I am getting SIGSEGV(11). Here is the cesm.log file:
`/glade/scratch/gmarques/g.e23_b16.GJRAv4.TL319_t232_zstar_N65.enthalpy_case5/run/cesm.log.3391111.chadmin1.ib0.cheyenne.ucar.edu.230919-142440
`

jedwards4b:
@gustavo-marques that issue should be fixed - update cmeps and try again.

gustavo-marques:
Just confirming I was able to run a G compset

Issue #267, opened 2022-02-01
Add convective gustiness to (ocean) surface flux calculation 


megandevlan:
We're working on a way to modify the calculation of surface fluxes (over ocean first) to include the impacts of convective gustiness. A few years ago, @swrneale implemented the modifications needed to do this in a branch of CAM, which mostly involved modifications to a few cime scripts (`shr_flux_mod.F90` and `seq_flux_mct.F90`) to add a gustiness term to the calculation of the surface wind magnitude `vmag` based on an input convective precipitation estimate from CAM. 

In the new CMEPS architecture, it seems this will involve modifying vmag in `cmeps/cesm/flux_atmocn/shr_flux_mod.F90` in the case where ocn_surface_flux_scheme=0. I suspect the (nearest?) equivalent to `seq_flux_mct.F90` is `cmeps/mediator/med_phases_aofluxes_mod.F90`, though I imagine a few other modifications might be necessary in `esmFldsExchange_cesm_mod.F90, fd_cesm.yaml` to pass the convective precipitation to the coupler.

PR #422, opened 2023-12-20
Add convective gust contribution to U10

megandevlan:
### Description of changes
Modifications made to enhance U10 computed in shr_flux_mod based on convective rain rate from CAM. Additional modifications made to output the magnitude of that enhancement (UGUST) in CAM as well. The option to turn on the enhancement is controlled by a name list variable, add_gusts. 

### Specific notes

Contributors other than yourself, if any: @swrneale

CMEPS Issues Fixed (include github issue #): https://github.com/ESCOMP/CMEPS/issues/267

Are changes expected to change answers? (specify if bfb, different at roundoff, more substantial) None if add_gusts=False

Any User Interface Changes (namelist or namelist defaults changes)? add_gusts option in namelist; default is false. 

### Testing performed

Tests on Derecho as part of most recent cam-dev: https://github.com/NCAR/amwg_dev/issues/440

### DIFF
--- a/cesm/flux_atmocn/shr_flux_mod.F90
+++ b/cesm/flux_atmocn/shr_flux_mod.F90
@@ -259,7 +259,17 @@ contains
 
 
     qsat(Tk)   = 640380.0_R8 / exp(5107.4_R8/Tk)
-    cdn(Umps)  =   0.0027_R8 / Umps + 0.000142_R8 + 0.0000764_R8 * Umps
+
+    ! Large and Yeager 2009
+    cdn(Umps)  =  0.0027_R8 / min(33.0000_R8,Umps) + 0.000142_R8 + &
+         0.0000764_R8 * min(33.0000_R8,Umps) - 3.14807e-13_r8 * min(33.0000_R8,Umps)**6
+    ! Capped Large and Pond by wind
+    !   cdn(Umps)  =   0.0027_R8 / min(30.0_R8,Umps) + 0.000142_R8 + 0.0000764_R8 * min(30.0_R8,Umps) 
+    ! Capped Large and Pond by Cd
+    !   cdn(Umps) = min(0.0025_R8, (0.0027_R8 / Umps + 0.000142_R8 + 0.0000764_R8 * Umps ))
+    ! Large and Pond
+    !   cdn(Umps)  =   0.0027_R8 / Umps + 0.000142_R8 + 0.0000764_R8 * Umps
+
     psimhu(xd) = log((1.0_R8+xd*(2.0_R8+xd))*(1.0_R8+xd*xd)/8.0_R8) - 2.0_R8*atan(xd) + 1.571_R8
     psixhu(xd) = 2.0_R8 * log((1.0_R8 + xd*xd)/2.0_R8)
 
--- a/cesm/flux_atmocn/shr_flux_mod.F90
+++ b/cesm/flux_atmocn/shr_flux_mod.F90
@@ -133,7 +133,7 @@ contains
   !                   Thomas Toniazzo (Bjerknes Centre, Bergen) ”
   !===============================================================================
   SUBROUTINE flux_atmOcn(logunit, nMax  ,zbot  ,ubot  ,vbot  ,thbot ,   &
-       &               qbot  ,s16O  ,sHDO  ,s18O  ,rbot,  &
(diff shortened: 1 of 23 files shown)


cesm/flux_atmocn/shr_flux_mod.F90
```diff
@@ -257,22 +263,21 @@ SUBROUTINE flux_atmOcn(logunit, nMax  ,zbot  ,ubot  ,vbot  ,thbot ,   &
     real(R8)    :: tdiff(nMax)               ! tbot - ts
     real(R8)    :: vscl
 
+    real(R8)    :: ugust      ! function: gustiness as a function of convective rainfall.  
+    real(R8)    :: gprec   ! dummy arg ~ ?
```
jedwards4b, line 267: Not a great comment here.

cesm/flux_atmocn/shr_flux_mod.F90
```diff
@@ -257,22 +263,21 @@ SUBROUTINE flux_atmOcn(logunit, nMax  ,zbot  ,ubot  ,vbot  ,thbot ,   &
...
-    cdn(Umps)  =  0.0027_R8 / min(33.0000_R8,Umps) + 0.000142_R8 + &
-         0.0000764_R8 * min(33.0000_R8,Umps) - 3.14807e-13_r8 * min(33.0000_R8,Umps)**6
-    ! Capped Large and Pond by wind
-    !   cdn(Umps)  =   0.0027_R8 / min(30.0_R8,Umps) + 0.000142_R8 + 0.0000764_R8 * min(30.0_R8,Umps) 
-    ! Capped Large and Pond by Cd
-    !   cdn(Umps) = min(0.0025_R8, (0.0027_R8 / Umps + 0.000142_R8 + 0.0000764_R8 * Umps ))
-    ! Large and Pond
-    !   cdn(Umps)  =   0.0027_R8 / Umps + 0.000142_R8 + 0.0000764_R8 * Umps
-
+    cdn(Umps)  =   0.0027_R8 / Umps + 0.000142_R8 + 0.0000764_R8 * Umps
```
jedwards4b, line 270: You removed the cap on this equation so this will affect not just cam_dev but all configurations?

jedwards4b (changes requested):
The cdn variable will change more than just cam_dev - can you clarify this is what you want.   Currently the add_gust variable is set to false and the only way to change it is in user_nl_cam or in a testmod.   I think you can set an esmf info object in cam to the result of cam_physpkg_is("cam_dev") then query that info object in cmeps to determine if the add_gust variable should be used.

cesm/flux_atmocn/shr_flux_mod.F90
megandevlan: Thanks Jim, fixed this now; not something I intended to change.

cesm/flux_atmocn/shr_flux_mod.F90
megandevlan: Updated this too; working on the esmf info object now.

megandevlan:
I'd been setting the namelist add_gust in usr_nl_cpl instead of usr_nl_cam; is the latter better practice?

jedwards4b:
No, I think that user_nl_cpl is the correct place, I was mistaken in my previous comment.

megandevlan:
Thanks Jim. I'm somewhat stumped for how to set an esmf object in CAM that can then be queried in ESMF; do you have an example of something similar being done that I could follow for some other variable? I'm guessing it's something that would be in the atm_comp_nuopc file in CAM?

jedwards4b:
@megandevlan I'm looking for the best solution and will get back to you.

jedwards4b:
@megandevlan I've opened [a PR](https://github.com/ESCOMP/CMEPS/pull/423) to your fork that addresses the cam-dev issue, since cam_dev is a compile time option and not a run time option we can do this check in the cmeps buildnml.   The PR also fixes the .github testing, so once it's merged all tests should pass.

megandevlan:
Thanks @jedwards4b, especially for getting to this so quickly!! These fixes looks great to me.

jedwards4b (approved):

cacraigucar:
@jedwards4b - Is there a tag for this change?

jedwards4b:
cmeps0.14.48  Tags are automatically generated in this repo, do a git fetch to see new tags.

jedwards4b:
cmeps0.14.49 now.

PR #425, opened 2023-12-27
For T compsets, only run GLC if med_to_glc is true

billsacks:
### Description of changes

This makes the logic for running GLC consistent with runseq_general.py. This change is relevant for the unusual case where a T compset (CISM forced by dlnd) is being run with CISM in noevolve mode: in this case, we don't want to run CISM, and trying to run CISM leads to incorrect behavior in a restart case (which is how this issue was detected).

With this change,
ERS_D_Ly3.f09_g17_gris4.T1850Gg.derecho_intel.cism-noevolve passes.

### Specific notes

Contributors other than yourself, if any: @Katetc 

CMEPS Issues Fixed (include github issue #): none

Are changes expected to change answers? No - bfb (except for the unusual case of a T compset with noevolve CISM)

### Testing performed

ERS_D_Ly3.f09_g17_gris4.T1850Gg.derecho_intel.cism-noevolve

Using escomp/cism-wrapper@bf0b382

### DIFF
--- a/cime_config/runseq/runseq_TG.py
+++ b/cime_config/runseq/runseq_TG.py
@@ -34,7 +34,7 @@ def gen_runseq(case, coupling_times):
         runseq.add_action ("MED med_phases_post_lnd"        , run_lnd)
         runseq.add_action ("MED med_phases_prep_glc"        , med_to_glc)
         runseq.add_action ("MED -> GLC :remapMethod=redist" , med_to_glc)
-        runseq.add_action ("GLC"                            , run_glc)
+        runseq.add_action ("GLC"                            , run_glc and med_to_glc)
         runseq.add_action ("GLC -> MED :remapMethod=redist" , run_glc)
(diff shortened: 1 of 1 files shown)


billsacks:
@Katetc can you please confirm that this solves your issue with `ERS_D_Ly3.f09_g17_gris4.T1850Gg.derecho_intel.cism-noevolve`? And can you also confirm that a normal (not noevolve) T compset still passes and is bit-for-bit with this change?

@jedwards4b I'm assigning you in order to have an official CMEPS gatekeeper as the official assignee.

jedwards4b (approved):

Katetc:
Hi everybody, this code update fixes the test in my sandbox. Thanks for your help @billsacks and @jedwards4b !

billsacks:
Thanks a lot @Katetc !

@jedwards4b I think this is ready to merge whenever you feel is a good time.

PR #429, opened 2024-02-14
[WIP] Add shr_dust_emis_mod to handle options for dust emissions between CTSM and CAM

### LINKED ISSUE #353: Add dust_emission_method to drv_flds_in
We need to add a new field to drv_flds_in that will be used by both CTSM and CAM. There is current work at adding a new dust emission scheme to CTSM. Both CTSM and CAM will need to know which scheme is being used. CTSM will toggle between the schemes and CAM needs to know which scheme is used in order to set the dust_emis_fact dust tuning factor.

Currently the only option is the Zender scheme, but Leung-2022 will be added as will a prescribed dust emissions mode.

Similar to other fields this will be able to be set by either CAM or CTSM with an error occurring if they disagree with each other. The field will then get passed down to both ATM and LND from the driver.

This was worked out in a meeting with @fvitt and others.

Definition of Done:
- [x] New FORTRAN code shr_dust_emis_mod.F90 in cesm/nuopc_cap_share
- [x] Update namelist defaults for new namelist items: dust_emission_method and zender_soil_erod_source

ekluzek:
### Description of changes

Add a shared module for controlling how dust emissions are handled in CTSM and CAM.
This will allow us to use the current dust emission options, and also the new dust emissions
being developed.

This implements the design given for this work in:

https://docs.google.com/document/d/18nZ3LJF5W-YF9iBhqed6s_NWeKOvSSL2-k0Lye1nnLg/edit#heading=h.c1sjcqnqh1zs

### Specific notes

Contributors other than yourself, if any: @fvitt 

  Fixes #353

Are changes expected to change answers? No

  Namelist items added the drv_fls_in namelist:

```
   dust_emis_method
   zender_soil_erod_source
```
   These will be set by either CAM or CTSM to control how dust emissions work.

### Testing performed

 None so far. But, will do standard testing in CTSM as well as any CMEPS test lists.
 This shouldn't effect anything outside of CTSM and CAM, and even then only for
 versions that use it.

### DIFF
--- a/cesm/nuopc_cap_share/shr_dust_emis_mod.F90
+++ b/cesm/nuopc_cap_share/shr_dust_emis_mod.F90
@@ -1,0 +1,204 @@
+module shr_dust_emis_mod
+
+  !========================================================================
+  ! Module for handling dust emissions.
+  ! This module is shared by land and atmosphere models for the computation of
+  ! dust emissions.
+  !========================================================================
+
+  use ESMF           , only : ESMF_VMGetCurrent, ESMF_VM, ESMF_VMGet, ESMF_LOGMSG_INFO
+  use ESMF           , only : ESMF_LogFoundError, ESMF_LOGERR_PASSTHRU, ESMF_SUCCESS
+  use ESMF           , only : ESMF_LogWrite, ESMF_VMBroadCast
+  use shr_sys_mod    , only : shr_sys_abort
+  use shr_kind_mod   , only : CS => SHR_KIND_CS
+  use shr_nl_mod     , only : shr_nl_find_group_name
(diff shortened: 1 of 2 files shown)


ekluzek:
@fvitt I haven't got this working or tested yet in CTSM. But, I think this includes everything needed in this space. So if you could look it over and give feedback that would be great. I'll be working on completing this to bring this in the next month or so...

fvitt (approved):
Looks good to me.
Thanks

ekluzek:
This is ready to come in from my perspective at this point. I'm happy to hear about any changes or testing that I need to do with it though.

And I figure this will have to wait until after the cesm3_0_beta01 freeze in case something is needed from CMEPS for the git fleximod update.

PR #444, opened 2024-03-27
fixes an issue with symbols in compset names on derecho

jedwards4b:
### Description of changes
Fixes an issue when running create_test with a full compset name that contains special characters like %.

### Specific notes
I changed the call to the Makefile from buildexe to use a relative path instead of full path, this seems to 
have addressed the issue.

### Testing performed


briandobbins (approved):
It's odd that we don't know the direct connection with GNU make and special characters (I still suspect the locale settings), but if this works in practice, passing our tests, that works for me.

PR #446, opened 2024-03-29
Fix/add gusts

jedwards4b:
### Description of changes
When coldair_outbreak_mod is true the vmax field was changed without a corresponding change in wind0.  This 
leads to an unexpected difference in the So_u10 field.  

### Specific notes

Contributors other than yourself, if any: @mvertens found the problem in her tests for dglc. 

Are changes expected to change answers? Corrects So_u10n

### Testing performed
SMS_Ln5.f19_f19_mg17.2000_DATM%QIA_SLND_SICE_DOCN%SOMAQP_SROF_SGLC_SWAV.derecho_intel
as compared to baseline in cdeps1.0.28

### DIFF
--- a/cesm/flux_atmocn/shr_flux_mod.F90
+++ b/cesm/flux_atmocn/shr_flux_mod.F90
@@ -347,7 +347,6 @@ contains
              !--- compute some needed quantities ---
              if (add_gusts) then 
                 vmag   = max(seq_flux_atmocn_minwind, sqrt( (ubot(n)-us(n))**2 + (vbot(n)-vs(n))**2 + (1.0_R8*ugust(min(rainc(n),6.94444e-4_r8))**2)) )
-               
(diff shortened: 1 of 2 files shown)


megandevlan:
I'd be curious to hear from @swrneale on this too. My instinct is that this should be the correct way of doing things, but does require a bit of a change in the way we've thought about this. 

Originally, I had thought of and defined `u10res` to be _only_ the large-scale resolved wind field, which should be based on UBOT and VBOT (as it is in the current code, stemming from `wind0`). But clearly, other parameterization options (coldair_outbreak_mod, in this case) can also impact the magnitude of the low-level wind (vmag). My thinking is ultimately that we want U10 as output from CMEPS to remain unchanged relative to before gusts were added, which requires that wind0 and u10res reflect all the ways in which vmag can change, including by the coldair_outbreak_mod (yes?). In that case, I agree that we need this change.

mvertens (approved):

PR #447, opened 2024-03-29
alarms based on nsteps 

### LINKED ISSUE #443: REST_OPTION=nsteps not working correctly
I don't believe that the mediator is correctly triggering RESTART writes.  In the following case I get restart files for all components except rof and cpl.  
```
create_newcase --case f.cam6_3_144.ne30.datest --res ne30pg3_ne30pg3_mg17 --compset FMTHIST --run-unsupported
./xmlchange REST_OPTION=nsteps
./xmlchange REST_N=12
./xmlchange STOP_OPTION=nsteps 
./xmlchange STOP_N=24 
./xmlchange DOUT_S=FALSE
```
I think that the issue is in the way that the restart timestep is calculated, from the mediator log I see:
`(med_phases_restart_alarm_init) restart clock timestep =      10800`
Which I believe is based on the ROF_NCPL=8.  We don't see this issue in the tests because the tests reset
ROF_NCPL=48 prior to running to allow the test to have a shorter overall length.

jedwards4b:
### Description of changes
Fixes the case of alarms based on nsteps when components have different individual timesteps.

### Specific notes

Contributors other than yourself, if any: @mvertens - Mariana was essential to this work, I was trying to work in the Fortran when she pointed out that we could avoid the problem by converting nsteps to another unit (seconds) in the python.   This made it much easier.

CMEPS Issues Fixed #443 

### Testing performed

The test was to run  ./create_newcase --case /glade/derecho/scratch/jedwards/f.cam6_3_144.ne30.datest --res ne30pg3_ne30pg3_mg17 --compset FMTHIST --run-unsupported
then set 
```
./xmlchange REST_OPTION=nsteps  
./xmlchange REST_N=12 
./xmlchange STOP_OPTION=nsteps
./xmlchange STOP_N=24 
```
Prior to this change the cpl and mosart restart files are not written at the proper time.   

This change should not affect any other alarm settings, only those using nsteps.  

@DeniseWorthen I have used nuopc_shr_methods.F90 in the mediator for this change, unfortunately this file is not shared with UFS.  We need to discuss how to rearrange the code so that this works for UFS models.


jedwards4b:
The proposed fix for this is to move nuopc_shr_methods.F90 into [cdeps/share](https://github.com/ESCOMP/CDEPS/pull/269) and
[cesm_share/src](https://github.com/ESCOMP/CESM_share/pull/47)

DeniseWorthen:
@jedwards4b If I understand, we already utilize some of the modules w/in CDEPS/share when we build CMEPS. So I think adding this nuopc_shr_method in that location would work.

jedwards4b:
@DeniseWorthen One thing that I wasn't sure of is if you use the same *_cpl_dt variables as cesm does.  If not we will need to figure out how to do this for ufs.

DeniseWorthen:
@jedwards4b I see. No, we don't use the cpl_dt variables in the same way. 

For restarts, ATM controls it's own restart writing frequency and we just ensure that the coupled components (using restart_n, restart_option) are in sync w/ that. We don't currently have any use case where we use nsteps as the restart frequency. 

But, unrelated, this comes along at the same time that I need to make sure we can always write restarts at the end-of-run (this is for when we use IAU). For some reason, I have in mind that the write_restart_at_endofrun option was not working.

mediator/med_time_mod.F90
```diff
@@ -180,9 +196,16 @@ subroutine med_time_alarmInit( clock, alarm, option, &
       update_nextalarm  = .false.
 
    case (optNSteps,trim(optNSteps)//'s')
-      call ESMF_ClockGet(clock, TimeStep=AlarmInterval, rc=rc)
+      call ESMF_ClockGet(clock, TimeStep=TimestepInterval, rc=rc)
+      if (ChkErr(rc,__LINE__,u_FILE_u)) return
+      call ESMF_TimeIntervalSet(AlarmInterval, s=min_timestep, rc=rc )
       if (ChkErr(rc,__LINE__,u_FILE_u)) return
       AlarmInterval = AlarmInterval * opt_n
+      if (mod(AlarmInterval, TimestepInterval) /= (timestepinterval*0)) then
```
mvertens, line 204: Please write a comment here as to why you are multiplying by 0.

cesm/driver/esm_time_mod.F90
```diff
@@ -54,7 +54,7 @@ module esm_time_mod
 !===============================================================================
 
   subroutine esm_time_clockInit(ensemble_driver, instance_driver, logunit, maintask, rc)
-
+    use nuopc_shr_methods, only : get_minimum_timestep, dtime_drv
```
mvertens, line 57: Comment that nuopc_shr_methods is now in the cdeps code and cdeps is always built alongside of cmeps - so this is available to cmeps.

mvertens (approved):

jedwards4b:
@DeniseWorthen I've added more error checking so that if someone in UFS does try to use the nsteps option it will throw an error.   I also tested REST_OPTION='end' and fixed an issue, but it seems that issue would only affect cesm - the fix was in the driver code.  Maybe the ufs driver is similar and needs the same change, but I don't have access to that code.

DeniseWorthen:
@jedwards4b I'm having trouble w/ the gust variables again. I've made this change
```
index ad2eda90..1678238d 100644
--- a/mediator/med_phases_aofluxes_mod.F90
+++ b/mediator/med_phases_aofluxes_mod.F90
@@ -1717,10 +1717,6 @@ end subroutine med_aofluxes_map_ogrid2xgrid_input
     call fldbun_getfldptr(fldbun, 'So_duu10n', aoflux_out%duu10n, xgrid=xgrid, rc=rc)

-    call fldbun_getfldptr(fldbun, 'So_ugustOut', aoflux_out%ugust_out, xgrid=xgrid, rc=rc)
-    if (chkerr(rc,__LINE__,u_FILE_u)) return
-    call fldbun_getfldptr(fldbun, 'So_u10withGust', aoflux_out%u10_withGust, xgrid=xgrid, rc=rc)
-    if (chkerr(rc,__LINE__,u_FILE_u)) return
     call fldbun_getfldptr(fldbun, 'So_u10res', aoflux_out%u10res, xgrid=xgrid, rc=rc)
     call fldbun_getfldptr(fldbun, 'Faox_taux', aoflux_out%taux, xgrid=xgrid, rc=rc)
@@ -1752,8 +1748,12 @@ end subroutine med_aofluxes_map_ogrid2xgrid_input
     if (add_gusts) then
        call fldbun_getfldptr(fldbun, 'So_ugustOut', aoflux_out%ugust_out, xgrid=xgrid, rc=rc)
+       call fldbun_getfldptr(fldbun, 'So_u10withGust', aoflux_out%u10_withGust, xgrid=xgrid, rc=rc)
+       if (chkerr(rc,__LINE__,u_FILE_u)) return
+
     else
        allocate(aoflux_out%ugust_out(lsize)); aoflux_out%ugust_out(:) = 0._R8
+       allocate(aoflux_out%u10_withGust(lsize)); aoflux_out%u10_withGust(:) = 0._R8
     end if

   end subroutine set_aoflux_out_pointers
```

Which I think I can fix. It seems the last gust PR created a duplicate line for getting the ``So_ugustOut`` fldptr. But I think putting them both inside the ``use_gusts`` works. 

I'm not sure what to do with the ``So_u10res`` pointer. Is that always used regardless of whether gusts are included or not?

jedwards4b:
@DeniseWorthen Your latest comment does not seem to be related to this PR.   Can you remove it and put it in the right place?  Did you mean to put it here? https://github.com/ESCOMP/CMEPS/pull/446

jedwards4b:
Adding code to fix REST_OPTION=end turns out to be a can of worms - for this to work stop_alarm must be set first and I'm finding that that is not the case in at least some of the nuopc component caps.  ugh

DeniseWorthen:
@jedwards4b I don't think the issue is related to #446. The current main contains 
https://github.com/ESCOMP/CMEPS/blob/a077536ccdf97e773a3fc586f90f559f9bc13edf/mediator/med_phases_aofluxes_mod.F90#L1720

and then a few lines down

https://github.com/ESCOMP/CMEPS/blob/a077536ccdf97e773a3fc586f90f559f9bc13edf/mediator/med_phases_aofluxes_mod.F90#L1752-L1757

This was done by @uturuncoglu in d56c50c. We now have another variable ``So_u10withGust`` which I thought could use the same fix (place inside of ``use_gusts``). But now there seems to be a third variable (``So_u10res``) and I'm not sure how to fix that one for UFS. The error I see is 

```
20240401 155101.781 ERROR            PET00 ESMCI_Container.h:178 ESMCI::Container::get() Invalid argument  - key does not exist: So_u10res
20240401 155101.781 ERROR            PET00 ESMCI_Container_F.C:451 c_esmc_containergetfield() Invalid argument  - Internal subroutine call returned Error
20240401 155101.781 ERROR            PET00 ESMF_Container.F90:636 ESMF_ContainerGetField() Invalid argument  - Internal subroutine call returned Error
20240401 155101.781 ERROR            PET00 ESMF_FieldBundle.F90:11581 ESMF_FieldBundleGetItem() Invalid argument  - Internal subroutine call returned Error
```

DeniseWorthen:
@jedwards4b I can't test your branch because it contains the changes related to gusts that are raising an issue on my side

https://github.com/jedwards4b/CMEPS/blob/44cb9019a00ff02191802be5bcccc7d9d4087d38/mediator/med_phases_aofluxes_mod.F90#L1720

jedwards4b:
but that change was not introduced here - find the pr that introduced the problem or open a new issue.

DeniseWorthen:
@jedwards4b Yes, thanks. I was trying to be responsive and test your nsteps PR. I am not able to test because the previous PR (which I will find) was merged w/o testing by UFS. So this is the first I've seen this failure. I'm sorry if I've been unclear.

DeniseWorthen:
@jedwards4b I've had to put this aside for a couple of days while I work on finishing something. But I will get back to it soon.

jedwards4b:
okay - I plan to work on it more anyway when derecho comes back.  There are three distinct copies of subroutine alarmInit in cmeps - I plan to merge these into a single routine, I also figured out a better solution to the REST_OPTION='end' issue.

jedwards4b:
@DeniseWorthen can you try this PR again and see if it works for you now.

DeniseWorthen:
Yes, will do. I'm not able to get onto Derecho right now though. It's still down for me.

jedwards4b:
yes, it's still down for everyone - i didn't realize that you were testing there too.

jedwards4b:
@DeniseWorthen I've updated to the latest cmeps main and tested this with cesm - do you want to try it again?

DeniseWorthen:
@jedwards4b Sure. I'll start the tests later today.

DeniseWorthen:
@jedwards4b I can't compile because it can't find ``nuopc_shr_methods``.

jedwards4b:
Sorry - can you try with this CDEPS branch:
To https://github.com/jedwards4b/CDEPS
 * [new branch]      add/nuopc_shr_methods.F90 -> add/nuopc_shr_methods.F90

DeniseWorthen:
Thanks. I think remember now that I did this before in testing. I can compile it now, so I'll start the tests.

DeniseWorthen:
@jedwards4b This doesn't cause any issues. I do get a new field in the CDEPS restart files (strlen) but that is the only impact.

cime_config/namelist_definition_drv.xml
```diff
@@ -233,17 +233,15 @@
...
+  <!-- not yet fully implemented, set to false -->
   <entry id="write_restart_at_endofrun">
     <type>logical</type>
     <category>nuopc</category>
     <group>ALLCOMP_attributes</group>
     <values>
       <value>.false.</value>
-      <value rest_option="end">.true.</value>
+      <value rest_option='none'>.false.</value>
+      <value rest_option='never'>.false.</value>
```
mvertens, line 244: Why did the rest_option="end" get removed?

cime_config/runseq/driver_config.py
```diff
@@ -77,15 +77,7 @@ def __compute_glc(self, case, coupling_times):
...
-        elif (comp_glc == 'dglc'):
-            glc_coupling_time = coupling_times["glc_cpl_dt"]
-            is_test = case.get_value("TEST")
-            if not is_test:
-                stop_option = case.get_value('STOP_OPTION')
-                if stop_option == 'nsteps':
-                    stop_n = case.get_value('STOP_N')
-                    glc_coupling_time = stop_n*coupling_times["atm_cpl_dt"]
-        elif (comp_glc == 'xglc'):
+        elif comp_glc == 'dglc' or comp_glc == 'xglc':
```
mvertens, line 80: Thanks. This is much cleaner this way.

mediator/med_phases_aofluxes_mod.F90
```diff
@@ -1753,8 +1750,11 @@ subroutine set_aoflux_out_pointers(fldbun, lsize, aoflux_out, xgrid, rc)
     if (add_gusts) then
        call fldbun_getfldptr(fldbun, 'So_ugustOut', aoflux_out%ugust_out, xgrid=xgrid, rc=rc)
        if (chkerr(rc,__LINE__,u_FILE_u)) return
+       call fldbun_getfldptr(fldbun, 'So_u10withGust', aoflux_out%u10_withGust, xgrid=xgrid, rc=rc)
```
mvertens, line 1753: This looks like it belongs in another PR associated with fluxes. Why is it in the alarms PR?

mvertens (changes requested):
I think it would be fewer changes to have med_time_mod.F90 simply use nuopc_shr_methods directly and not get rid of the file. That way it becomes explicit as to what is happening and there are much fewer changes to this code.

cime_config/namelist_definition_drv.xml
jedwards4b: Because write_restart_at_endofrun still isn't working correctly.   I will address this in an upcoming PR.

mvertens (approved):

PR #449, opened 2024-04-02
remove  ESMF_GridCompGetInternalState from use statement, not working…

jedwards4b:
… with nag and apparently not needed

### Description of changes
Remove the use ESMF_GridCompGetInternalState it is not needed and fails for the nag compiler.  

### Specific notes

### Testing performed

### DIFF
--- a/mediator/med_phases_aofluxes_mod.F90
+++ b/mediator/med_phases_aofluxes_mod.F90
@@ -1601,10 +1601,8 @@ end subroutine med_aofluxes_map_ogrid2xgrid_input
        if (chkerr(rc,__LINE__,u_FILE_u)) return
        call fldbun_getfldptr(fldbun_a, 'Sa_shum', aoflux_in%shum, xgrid=xgrid, rc=rc)
        if (chkerr(rc,__LINE__,u_FILE_u)) return
-       if (add_gusts) then
-          call fldbun_getfldptr(fldbun_a, 'Faxa_rainc', aoflux_in%rainc, xgrid=xgrid, rc=rc)
-          if (chkerr(rc,__LINE__,u_FILE_u)) return
-       end if
+       call fldbun_getfldptr(fldbun_a, 'Faxa_rainc', aoflux_in%rainc, xgrid=xgrid, rc=rc)
+       if (chkerr(rc,__LINE__,u_FILE_u)) return
     end if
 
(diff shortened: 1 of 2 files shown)


mediator/med_phases_aofluxes_mod.F90
```diff
@@ -1601,10 +1601,8 @@ subroutine set_aoflux_in_pointers(fldbun_a, fldbun_o, aoflux_in, lsize, xgrid, r
        if (chkerr(rc,__LINE__,u_FILE_u)) return
        call fldbun_getfldptr(fldbun_a, 'Sa_shum', aoflux_in%shum, xgrid=xgrid, rc=rc)
        if (chkerr(rc,__LINE__,u_FILE_u)) return
-       if (add_gusts) then
```
jedwards4b, line 1604: @ekluzek I wish to point out this change required for nag because it doesn't like the unallocated pointer in the subroutine interface even though the pointer is never used.   This is a case where the nag compiler is requiring changes we don't otherwise need to make.

jedwards4b:
I ran and passed  ERC_D_Ln9.T5_T5_mg37.QPC4.izumi_nag.cam-outfrq3s_usecase, due to the urgency of the request this code change was not reviewed prior to merging.

billsacks (commented):
Approving after the fact... thank you very much @jedwards4b !

mediator/med_phases_aofluxes_mod.F90
billsacks: I see arguments both ways here, but I lean slightly towards saying that this is an argument *for* keeping nag: I believe this is enforcing an aspect of the Fortran standard, so by fixing this we (1) make it more likely that our code won't be rejected by various other compilers that we don't currently test with, and (2) avoid the potential for undefined behavior in compilers that accept the code.

ekluzek: My main argument for nag itself is that it's found legit subscript overflow problems that for whatever reasons other compilers haven't caught. If we can use different options to catch those problems I'm happy to move away from nag.

ekluzek: In terms of the code here, I see Jim's point that it's annoying here. But, if Bill is right that it's enforcing the standard I do side towards that. But, it also might be something that other compilers could find by using stricter options to enforce the standard?

PR #460, opened 2024-05-17
avoid integer overflow by skipping mediator in restart loop

jedwards4b:
### Description of changes
This loop in med_phases_restart_mod.F90 was causing integer overflow because the med component should not be included.   This was only happening with the nag compiler in debug mode and would not normally be a problem since integer overflows are ignored (except by nag in debug mode) and the resulting values were never used.  

### Specific notes

### Testing performed

### DIFF
--- a/mediator/med.F90
+++ b/mediator/med.F90
@@ -2122,7 +2122,7 @@ contains
        do n1 = 1,ncomps
           if (maintask) then
              write(logunit,*)
-             write(logunit,'(a)') trim(subname)//" "//trim(compname(n1))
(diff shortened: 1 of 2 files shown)


mvertens:
@jedwards - what testing was done for this? Can you please run a few tests to demonstrate that this fix does not adversely effect other configurations.

mvertens (approved):
This looks fine. It would be good to add a comment as to why you are ignoring this for the mediator.

ekluzek:
We have a long lists of tests that fail in our CTSM PR https://github.com/ESCOMP/CTSM/pull/2539. It would be good to have this come in as a tag so we can show those tests PASS.

What testing needs to be done for this to come in? Maybe the aux_cmeps and aux_clm test lists in a CESM checkout?

PR #481, opened 2024-07-09
Separate the control of removing negative runoff for lnd vs glc, and set default to false for glc-derived runoff

billsacks:
### Description of changes

This is a follow-up to #471 .

At least for now, we want the default to be to remove negative runoff for lnd-derived runoff, but NOT for glc-derived runoff. This PR implements that change. Note that this changes behavior for the correction of glc-derived runoff, but keeps the behavior the same as before for lnd-derived runoff.

### Specific notes

Contributors other than yourself, if any: General idea based on discussion with @mvertens , @gustavo-marques, @Katetc , @wwieder, @slevis-lmwg,  @whlipscomb, @gunterl

CMEPS Issues Fixed (include github issue #): None

Are changes expected to change answers? (specify if bfb, different at roundoff, more substantial): **Yes - changes answers substantially for runoff fields in some B compsets (or other configurations with both active land/river and active ocean) - but only for configurations that have glc-derived runoff; this can currently come from configurations with CISM in EVOLVE mode or DGLC (even in NOEVOLVE mode). Note that this change just undoes part of the answer changes that came from #471 .**

Any User Interface Changes (namelist or namelist defaults changes)? Separates the namelist item remove_negative_runoff into two separate options: remove_negative_runoff_lnd and remove_negative_runoff_glc.

### Testing performed

Using `cesm2_3_beta17` (with CISM at `756cfa6f0514d89977d2732ad06a4d364da5d418`, mosart at mosart1.1.02, and removing the contents of `cime_config/testmods_dirs/allactive/defaultio/user_nl_mosart`: Ran three versions of `SMS_D_Ld3.f19_g17.B1850G.derecho_intel.allactive-cism-test_coupling`:
- One with default flags
- One with `remove_negative_runoff_glc` set to true
- One with `remove_negative_runoff_lnd` set to false

I checked the behavior in these in runs with dbug_flag set to 21 so I would see diagnostics of what fields were being adjusted, and confirmed that the correct fields were being operated on in all three cases.

### DIFF
--- a/cime_config/namelist_definition_drv.xml
+++ b/cime_config/namelist_definition_drv.xml
@@ -872,17 +872,28 @@
       <value COMP_LND="xlnd">off</value>
     </values>
   </entry>
-  <entry id="remove_negative_runoff">
+  <entry id="remove_negative_runoff_lnd">
     <type>logical</type>
     <category>control</category>
     <group>MED_attributes</group>
     <desc>
-      If true, remove negative runoff by downweighting all positive runoff globally.
+      If true, remove negative runoff generated from the land component by downweighting all positive runoff globally.
     </desc>
     <values>
       <value>.true.</value>
     </values>
   </entry>
+  <entry id="remove_negative_runoff_glc">
+    <type>logical</type>
+    <category>control</category>
+    <group>MED_attributes</group>
+    <desc>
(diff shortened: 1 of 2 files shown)


mvertens (approved):
Looks good.

PR #484, opened 2024-07-11
fix bug in ocn2glc_coupling and add blom 

mvertens:
### Description of changes
This adds blom as another ocean component in cime_config.

### Specific notes
This adds blom as another ocean component in cime_config and also fixes the query of the logical variable for ocn2glc coupling in esmFldsExchange_cesm.F90.

CMEPS Issues Fixed: No

### Testing performed
Verified that ocn2glc coupling is working correctly with blom.

### DIFF
--- a/cime_config/config_component_cesm.xml
+++ b/cime_config/config_component_cesm.xml
@@ -226,14 +226,15 @@
       <!-- =================================================== -->
       <value compset="_DATM.*_DICE.*_POP2">24</value>
       <value compset="_DATM.*_DICE.*_MOM6">24</value>
-      <!-- TODO: currently a NUOPC runseqence cannot be generated with the ATM_NCPL < OCN_NCPL - as is the case -->
-      <!-- with the POP2 C compset run sequence - for now will set these to be the same as MOM6 -->
+      <value compset="_DATM.*_DICE.*_BLOM">24</value>
+      <!-- NOTE: urrently a NUOPC runseqence cannot be generated with the ATM_NCPL < OCN_NCPL -->
       <!-- =================================================== -->
       <!-- G compsets -->
       <!-- =================================================== -->
       <value compset="_DATM.*_SLND.*_CICE.*_POP2">24</value>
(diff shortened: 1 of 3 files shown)


cime_config/config_component_cesm.xml
```diff
@@ -190,21 +226,23 @@
       <!-- =================================================== -->
       <value compset="_DATM.*_DICE.*_POP2">24</value>
       <value compset="_DATM.*_DICE.*_MOM6">24</value>
-      <!-- TODO: currently a NUOPC runseqence cannot be generated with the ATM_NCPL < OCN_NCPL - as is the case -->
-      <!-- with the POP2 C compset run sequence - for now will set these to be the same as MOM6 -->
+      <value compset="_DATM.*_DICE.*_BLOM">24</value>
+      <!-- NOTE: urrently a NUOPC runseqence cannot be generated with the ATM_NCPL < OCN_NCPL -->
```
billsacks, line 230: Minor typo: urrently -> currently

cime_config/config_component_cesm.xml
```diff
@@ -141,6 +141,42 @@
...
+
+  <entry id="NH3_EMIS_OCN">
+    <type>logical</type>
+    <default_value>FALSE</default_value>
+    <valid_values>TRUE,FALSE</valid_values>
+    <group>run_coupling</group>
+    <file>env_run.xml</file>
+    <desc>Activates NH3 fluxes to be sent from ocn to atm </desc>
+  </entry>
+
```
billsacks, line 179: Are these only valid for BLOM currently? If so, I feel this should at least be documented. Ideally they would only be available for BLOM, or there would be an error check to ensure they aren't set for other models, but I suspect that's hard... if it's true that those would be hard to do, then I think it's fine to just add in the description something like "Only available with the BLOM ocean model" or something like that.

billsacks (changes requested):
A couple of questions / minor (I hope) requests below.

mvertens:
@billsacks - thanks so much for your review. I've addressed the issues you raised.

billsacks (approved):
Looks good - thank you!

PR #493, opened 2024-08-05
Add marbl

mnlevy1981:
### Description of changes

Update CESM component environment variables when turning MARBL tracers on changes expected default values.
Also removed default values that are based on having POP2 as the ocean component since it is no longer available.

### Specific notes

Want `CCSM_BGC=CO2A` and `CCSM_CO2_PPMV=284.317` when MOM6 is configured with MARBL tracers.

Are changes expected to change answers? This is technically bit-for-bit since it will not change answers unless MARBL is turned on and there are currently no tests with MARBL active.

### Testing performed

I've run `aux_mom` with these changes, and confirmed bfb in non-MARBL configurations.


mnlevy1981:
We can't make `cesm3_0_alpha03b` without this PR merged in, but it can go into `alpha03a` (it doesn't require any specific tags in `alpha03b`)

mnlevy1981:
With the change to the plans database, this is needed in `3_0_alpha03a` (the MOM6 tag with MARBL is coming in that tag, not `alpha03b`)

jedwards4b (approved):
LGTM

Issue #133, opened 2020-11-14
Simplify logic for glc_renormalize_smb


billsacks:
I think that this logic for glc_renormalize_smb in CMEPS:

https://github.com/ESCOMP/CMEPS/blob/0315cae1bc0891596f37dabbaf262a2f9267cc04/mediator/med_phases_prep_glc_mod.F90#L229-L247

differs from the equivalent logic in MCT:

https://github.com/ESMCI/cime/blob/9db73e464988bbc10779128b350f6681205696c7/src/drivers/mct/main/prep_glc_mod.F90#L334-L356

In particular, note that the MCT side has a flag `glc_coupled_fluxes`. This flag is set by CISM based on whether CISM is operating in a fully-coupled sense. The need for this arises from the possibility that CISM can be running in a mode where it is doesn't send liquid/ice fluxes to the ocean (instead relying on CTSM to send the fluxes needed for conservation). (This is always the case when ice evolution is turned off, but it can also be the case even when ice evolution is turned on, depending on how a flag is set.) When we first implemented the renormalization, land ice scientists felt that the default behavior should be for renormalization to be off in this case.

@mvertens if it's straightforward to set the `glc_coupled_fluxes` logical in CISM's nuopc_cap, similarly to what's done in the MCT cap, then this could be made consistent. Note that this `glc_coupled_fluxes` flag potentially differs for each ice sheet when running with multiple ice sheets. If there's some challenge with doing this, then we could check with @whlipscomb and others if it's acceptable to change the default behavior.

**Update (2021-10-18): See https://github.com/ESCOMP/CMEPS/issues/133#issuecomment-946219323 for the latest plan.**

billsacks:
For multiple ice sheets, @mvertens has made this work so that the "global" sums are actually just done over each ice sheet independently (this is accomplished by using a multiplying factor of icemask for the given ice sheet). I have verified that, with renormalization on, the results for Greenland are identical between a two-ice sheet run and a single ice sheet run, and similarly for Antarctica, so this seems to be working correctly (assuming you agree that this is the correct way to do the renormalization).

[...]

PR #495, opened 2024-08-09
Simplify logic for glc_renormalize_smb

billsacks:
### Description of changes

Remove the glc_renormalize_smb option on_if_glc_coupled_fluxes. This had been the default. Change the default to 'on'. This maintains consistent behavior between different configurations.

### Specific notes

CMEPS Issues Fixed (include github issue #): Resolves ESCOMP/CMEPS#133

Are changes expected to change answers? YES: changes answers by greater than roundoff for T compsets (active GLC with DLND) (now SMB is remapped with renormalization for T compsets by default); answers should be bit-for-bit in all other configurations (unless there is some configuration/combination that I'm not thinking of...)

Any User Interface Changes (namelist or namelist defaults changes)? Removes `on_if_glc_coupled_fluxes` option to `glc_renormalize_smb`

### Testing performed

Tested in the context of `cesm3_0_beta02`: ran:
- SMS_Ld5.f10_f10_ais8gris4_mg37.I1850Clm50SpGag.derecho_intel.cism-test_coupling
- SMS_Ld5.f10_f10_mg37.1850_DATM%GSWP3v1_CLM50%SP_SICE_SOCN_MOSART_DGLC%NOEVOLVE_SWAV.derecho_intel [Like an I compset but with DGLC%NOEVOLVE]
- SMS_Ly2.f09_g17_gris20.T1850Gg.derecho_intel

All tests passed; the T compset test changed answers as expected but the other two were bit-for-bit.

### DIFF
--- a/cime_config/buildnml
+++ b/cime_config/buildnml
@@ -98,7 +98,7 @@ def _create_drv_namelists(case, infile, confdir, nmlgen, files):
     else:
         config["single_column"] = "false"
 
-    # needed for determining the run sequence as well as glc_renormalize_smb
+    # needed for determining the run sequence
     config["COMP_ATM"] = case.get_value("COMP_ATM")
     config["COMP_ICE"] = case.get_value("COMP_ICE")
     config["COMP_GLC"] = case.get_value("COMP_GLC")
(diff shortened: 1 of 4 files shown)


billsacks:
@mvertens @whlipscomb @Katetc - this (finally) resolves #133 in the way we discussed there a few years ago. Note that this will change answers for T compsets, because now the SMB renormalization is always on by default (but it can still be turned off if desired).

mvertens (approved):
Thanks for this. Its so much simpler now!

PR #498, opened 2024-08-12
wav2ocn and ocn2wav maps are still needed in some cases

jedwards4b:
### Description of changes
Reintroduce wav2ocn and ocn2wav map xml variables.

### Specific notes

Contributors other than yourself, if any: mvertens

### Testing performed

### DIFF
--- a/cime_config/config_component.xml
+++ b/cime_config/config_component.xml
@@ -1405,6 +1405,22 @@
     <desc>rof2ocn runoff mapping file</desc>
   </entry>
 
+  <entry id="OCN2WAV_SMAPNAME">
+    <type>char</type>
+    <default_value>idmap</default_value>
+    <group>run_domain</group>
+    <file>env_run.xml</file>
+    <desc>ocn2wav state mapping file</desc>
+  </entry>
+
+  <entry id="WAV2OCN_SMAPNAME">
+    <type>char</type>
+    <default_value>idmap</default_value>
+    <group>run_domain</group>
+    <file>env_run.xml</file>
+    <desc>wav2ocn state mapping file</desc>
+  </entry>
+  
   <entry id="EPS_FRAC">
     <type>char</type>
     <default_value>1.0e-02</default_value>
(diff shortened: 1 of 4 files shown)


cime_config/config_component.xml
```diff
@@ -1405,6 +1405,22 @@
     <desc>rof2ocn runoff mapping file</desc>
   </entry>
 
+  <entry id="OCN2WAV_SMAPNAME">
```
mvertens, line 1408: I think you want something of the form
  <entry id="ocn2wav_map">
    <type>char</type>
    <category>mapping</category>
    <group>MED_attributes</group>
    <desc>ocn to wav mapping, 'unset' or 'idmap' are normal possible values</desc>
    <values>
      <value>unset</value>
      <value samegrid_ocn_wav='true'>idmap</value>
    </values>
  </entry>

And set samegrid_ocn_wav in buildnml. It doesn't make sense to have idmap be the default value if the two are on different grids.

cime_config/config_component.xml
```diff
@@ -1405,6 +1405,22 @@
...
+  </entry>
+
+  <entry id="WAV2OCN_SMAPNAME">
+    <type>char</type>
+    <default_value>idmap</default_value>
+    <group>run_domain</group>
+    <file>env_run.xml</file>
+    <desc>wav2ocn state mapping file</desc>
+  </entry>
+
```
mvertens, line 1423: Same comment as above.

cime_config/config_component.xml
mvertens: However, do we really want to have these as xml variables - or just add them in as namelist variables in case the user might want to change them?

mvertens (changes requested):

cime_config/config_component.xml
jedwards4b: I think maybe there is a format error in your comment - if you meant to include xml can you wrap it in <CODE></CODE>.
Also I don't see a variable samegrid_ocn_wav in the cmeps buildnml.   I'm not sure what you want.

alperaltuntas (approved):

mvertens (approved):

PR #506, opened 2024-09-28
Fix xgrid reproducibility by using srcTermProcessing=0 for all xgrid FieldRegridStore calls

### LINKED ISSUE #505: Runs with exchange grid give different results each time they're run
I noticed that runs with `aoflux_grid = "xgrid"` give different results each time they're run. I noticed this in cesm3_0_alpha03c with test `SMS_Ld2.ne30pg3_t232.BLT1850.derecho_intel.allactive-defaultio` when changing `aoflux_grid = "xgrid"`.

I have a fix for this; I'll submit a PR shortly.

billsacks:
### Description of changes

Use srcTermProcessing=0 for all xgrid FieldRegridStore calls.

This is needed for bit-for-bit reproducibility. Most of the calls to FieldRegridStore weren't setting srcTermProcessing at all; this can lead to irreproducibility of the results. Two of the calls had srcTermProcessing set to 1 before, which leads to reproducibility but (according to Gerhard Theurich) is often worse for performance than a value of 0.

Note that we use a value of 0 in the calls in med_map_mod.

### Specific notes

Contributors other than yourself, if any: Guidance from @oehmke and @theurich

CMEPS Issues Fixed (include github issue #): Resolves #505 

Are changes expected to change answers? YES: roundoff-level differences expected for runs with `aoflux_grid = "xgrid"` (but those runs had expected roundoff-level differences from run to run before this fix anyway)

### Testing performed

In the context of cesm3_0_alpha03c, with the change here put on top of the change in #501, ran `REP_Ld2.ne30pg3_t232.BLT1850.derecho_intel.allactive-xgrid`, where the `xgrid` testmod contained:

include_user_mods:
```
../defaultio
```

user_nl_cpl:
```
aoflux_grid = "xgrid"
```

### DIFF
--- a/mediator/med_phases_aofluxes_mod.F90
+++ b/mediator/med_phases_aofluxes_mod.F90
@@ -780,13 +780,15 @@ contains
     type(ESMF_Mesh)      :: xch_mesh
     real(r8), pointer    :: dataptr(:)
     integer              :: fieldcount
-    integer              :: stp ! srcTermProcessing is declared inout and must have variable not constant
+    integer              :: srcTermProcessing_Value ! srcTermProcessing is declared inout and must have variable not constant
     type(ESMF_CoordSys_Flag)           :: coordSys
     real(ESMF_KIND_R8)    ,allocatable :: garea(:)
     character(len=*),parameter :: subname=' (med_aofluxes_init_xgrid) '
     !-----------------------------------------------------------------------
 
     rc = ESMF_SUCCESS
+
+    srcTermProcessing_Value = 0
 
     ! Get the internal state from the mediator Component.
     nullify(is_local%wrap)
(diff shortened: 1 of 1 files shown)


mediator/med_phases_aofluxes_mod.F90
```diff
@@ -877,23 +879,26 @@ subroutine med_aofluxes_init_xgrid(gcomp, aoflux_in, aoflux_out, rc)
...
-    call ESMF_FieldRegridStore(xgrid, field_a, field_x, routehandle=rh_agrid2xgrid, rc=rc)
+    call ESMF_FieldRegridStore(xgrid, field_a, field_x, routehandle=rh_agrid2xgrid, &
+         srcTermProcessing=srcTermProcessing_Value, rc=rc)
     if (chkerr(rc,__LINE__,u_FILE_u)) return
     call ESMF_FieldRegridStore(xgrid, field_a, field_x, routehandle=rh_agrid2xgrid_2ndord, &
-         regridmethod=ESMF_REGRIDMETHOD_CONSERVE_2ND, rc=rc)
+         regridmethod=ESMF_REGRIDMETHOD_CONSERVE_2ND, srcTermProcessing=srcTermProcessing_Value, rc=rc)
     if (chkerr(rc,__LINE__,u_FILE_u)) return
     if (trim(coupling_mode) == 'cesm') then
-       stp = 1
```
billsacks, line 886: @jedwards4b - particularly note that I changed this value from 1 to 0 even for places where you were already setting srcTermProcessing. My sense is that you chose a value of 1 based on guidance from @oehmke (based on looking back at some email exchanges from last year), and in a conversation with @theurich and @oehmke yesterday they supported using 0 instead of 1. But let me know if you had reasons for choosing a value of 1 here that we should maintain. Note that we use a value of 0 in med_map_mod. According to @theurich, both 0 and 1 ensure bit-for-bit reproducibility, but he feels that 0 is probably better for performance in many cases.

billsacks:
I have run two additional REP tests (`REP_Ld2.ne30pg3_t232.BLT1850.derecho_intel.allactive-xgrid`) and verified that they pass as well, so I'm fairly confident that this fixes the reproducibility issue (at least for this configuration) (i.e., the original pass wasn't just by chance).

jedwards4b (approved):
Thanks!

PR #512, opened 2024-10-04
Update XML variables for GPU configurations

sjsprecious:
As discussed in https://github.com/ESMCI/cime/pull/4687, it is better remove the GPU options from the Python workflow in CIME and use XML files instead to configure them for CESM.

This PR introduces the necessary changes to accomplish this goal.

It should work together with the following PRs:
- CIME: https://github.com/ESMCI/cime/pull/4690
- ccs_config_cesm: https://github.com/ESMCI/ccs_config_cesm/pull/189

### DIFF
--- a/cime_config/config_component.xml
+++ b/cime_config/config_component.xml
@@ -795,20 +795,38 @@
 
   <entry id="GPU_TYPE">
     <type>char</type>
-    <valid_values></valid_values>
-    <default_value></default_value>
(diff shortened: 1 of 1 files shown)


sjsprecious:
@jedwards4b somehow I could not add you as a reviewer so I just mention you here.

jedwards4b (approved):

PR #515, opened 2024-10-21
Make xgrid the default aoflux_grid when atm and ocn are running on different grids

### LINKED ISSUE #502: Change CESM's default aoflux_grid to xgrid
We want to use the exchange grid by default for the aoflux_grid.

There have been some issues getting this to work in all cases, but we seem to be getting closer to having this working. Specifically:
- [ ] @oehmke has a fix for an issue with 2nd-order-conservative mapping with the exchange grid that can show up with certain grids (though not in CESM's current prealpha test suite). This should make it into the ESMF 8.8 development code base soon.
- [x] #501 fixes some CMEPS issues with add_gusts false
- [x] (Edit: Added 2024-09-30) #506 fixes xgrid reproducibility
- [x] (Edit: Added 2024-10-01) Resolve #510 
- [x] (Edit: Added 2024-10-18) Merge #514 

After those three pieces are ready and we have updated CESM to point to an ESMF development snapshot with the fix noted above, I think we should switch the default to aoflux_grid to "xgrid" so that we are testing with that configuration. @jedwards4b do you agree with this timing?

billsacks:
### Description of changes

Use the exchange grid for the atmosphere-ocean flux calculation unless atm & ocn are on the same grid.

Also adds a test of aoflux_grid=ogrid to the aux_cmeps test suite so that that configuration continues to be tested.

### Specific notes

Contributors other than yourself, if any: Consultation with @mvertens on how we should set the default – specifically, continuing to use ogrid when atm & ocn are on the same grid.

- Resolves #502 

Are changes expected to change answers? YES - significant answer changes for any configuration where atmosphere-ocean flux calculations are performed and atm & ocn are on different grids.

Any User Interface Changes (namelist or namelist defaults changes)? Changes default value for aoflux_grid

### Testing performed

Ran the full CESM prealpha and prebeta test suites on derecho using cesm3_0_beta03, but with the following commits cherry-picked into CMEPS:
- "Use xgrid for aoflux calculation unless atm & ocn are on the same grid" (from this PR)
- "Add a test with aoflux_grid = ogrid" (from this PR)
- "Ensure psfc is always allocated" (from #514)
- "More correct fix to ensure psfc is allocated when needed" (from #514)
- "Merge pull request #501 from billsacks/fix_xgrid_f_compset"
- "Merge pull request #506 from billsacks/fix_xgrid_reproducibility"

I compared my failures with failures in the baselines from @fischer-ncar 's testing. I got a number of additional failures due to tests exceeding walltime limits, which I'm ignoring (7 failures in prealpha and 7 failures in prebeta). I also got a failure in `PET_PM.ne30pg3_t232.BLT1850.derecho_intel.allactive-defaultiomi` and the similar gnu test, `PET_PM.ne30pg3_t232.BLT1850.derecho_gnu.allactive-defaultiomi`, with an error that seems unrelated to the xgrid: "CAM configure: ERROR: The SE dycore does not currently work with threading on." Other than those, tests passed.

### DIFF
--- a/cime_config/testdefs/testlist_drv.xml
+++ b/cime_config/testdefs/testlist_drv.xml
@@ -65,6 +65,16 @@
     </options>
   </test>
 
+  <test compset="X" grid="f19_g17" name="ERS_D_Ln9" testmods="drv/aoflux_ogrid">
+    <machines>
+      <machine name="derecho" compiler="intel" category="aux_cmeps"/>
+    </machines>
+    <options>
+      <option name="wallclock"> 00:40:00 </option>
+      <option name="comment">The default aoflux_grid is now xgrid; include a debug test with different atm/ocn grids with aoflux_grid=ogrid to make sure that option continues to work.</option>
+    </options>
+  </test>
+
   <!-- ======================================= -->
   <!-- B compsets -->
   <!-- ======================================= -->
--- a/cime_config/namelist_definition_drv.xml
+++ b/cime_config/namelist_definition_drv.xml
@@ -925,7 +925,10 @@
       default: ogrid
     </desc>
     <values>
-      <value>ogrid</value>
+      <value>xgrid</value>
+      <!-- xgrid introduces unnecessary cost and complexity if atm & ocn are on the same
+      grid, so fall back to ogrid in that case -->
+      <value samegrid_atm_ocn='true'>ogrid</value>
     </values>
   </entry>
   <entry id="ocn_surface_flux_scheme">
--- a/cime_config/testdefs/testmods_dirs/drv/aoflux_ogrid/user_nl_cpl
+++ b/cime_config/testdefs/testmods_dirs/drv/aoflux_ogrid/user_nl_cpl
@@ -1,0 +1,1 @@
+aoflux_grid = "ogrid"


billsacks:
@jedwards4b - based on discussion with Bob Oehmke, it might still be at least a couple more weeks before his exchange grid fixes make it into ESMF. I am therefore thinking that we should bring in the change to using xgrid by default (i.e., this PR) soon rather than waiting for the updated ESMF version. Does that make sense to you?

However, this should come in after #514 is merged.

cime_config/testdefs/testlist_drv.xml
```diff
@@ -65,6 +65,16 @@
     </options>
   </test>
 
+  <test compset="X" grid="f19_g17" name="ERS_D_Ln9" testmods="drv/aoflux_ogrid">
+    <machines>
+      <machine name="derecho" compiler="intel" category="aux_cmeps"/>
```
billsacks, line 70: I felt like it would be good to have at least one test that still covers doing the aoflux calculation on the ogrid to ensure that that configuration continues to work. It felt right to add this to aux_cmeps since this is about testing cmeps code, but I'm not sure how often aux_cmeps is actually run. @jedwards4b do you have feelings on whether aux_cmeps is reasonable or if we should add this to prealpha or prebeta (instead or in addition)?

Also, I wasn't positive if an X compset actually covers the atm-ocn flux calculation, and so if this X compset test is sufficient for exercising aoflux_grid=ogrid. @jedwards4b or @mvertens do you know? If you aren't sure, I can try to confirm that we're covering relevant code with this test, or just change it to a B compset.

billsacks:
> this should come in after https://github.com/ESCOMP/CMEPS/pull/514 is merged

That PR has been merged. So I feel like this can come in at any point. Outstanding questions are:
- Are we ready to bring in this change to xgrid being the default? (@jedwards4b )
- Is the added test here a reasonable one for continuing to test ogrid, both in terms of the test suite (aux_cmeps) and compset (X)? (@jedwards4b / @mvertens )

cime_config/testdefs/testlist_drv.xml
jedwards4b: I agree that retaining at least one ogrid case is a good idea.

jedwards4b (approved):
thank you!

jedwards4b:
I think maybe if you rebase it will fix the workflow error.

billsacks:
@jedwards4b - updating the PR to latest main didn't solve the workflow issue ("ModuleNotFoundError: No module named 'distutils'")

jedwards4b:
THis is fixed in the latest cime, updating to the latest cmeps should get the latest cime, anyway I understand what the issue is and it isn't your PR.

PR #522, opened 2024-12-18
Add arbitrary restart hour for UFS

DeniseWorthen:
### Description of changes

Enables UFS to read a list of non-interval restart hours and write restarts when appropriate. This is required for operational implementations. 

Incidentally adds new Land model component (LM4) for UFS and modifies CPP metadata for the FBH (fire-behavior model).

### Specific notes

Code is added under an CESMCOUPLED flag and accesses share code within the UFS CDEPS share code (ufs/cdeps_share). 

See related discussion https://github.com/orgs/esmf-org/discussions/291

@nickszapiro-noaa

https://github.com/ufs-community/ufs-weather-model/issues/2348

Should not impact CESM

### Testing performed


DeniseWorthen:
@jedwards4b I'm not sure, but I might need an ifdef around the restart_fh logical in ``med_phases_restart_write`` to avoid and unused variable warning on your side?

mediator/med_phases_restart_mod.F90
```diff
@@ -179,6 +189,7 @@ subroutine med_phases_restart_write(gcomp, rc)
     real(R8)                   :: tbnds(2)       ! CF1.0 time bounds
     logical                    :: isPresent
     logical                    :: first_time = .true.
+    logical                    :: write_restartfh
```
jedwards4b, line 192: Need an ifdef here.

mediator/med_phases_restart_mod.F90
```diff
@@ -14,6 +14,9 @@ module med_phases_restart_mod
   use med_phases_prep_glc_mod , only : FBocnAccum2glc_o, ocnAccum2glc_cnt
   use med_phases_prep_rof_mod , only : FBlndAccum2rof_l, lndAccum2rof_cnt
   use pio                     , only : file_desc_t
+#ifndef CESMCOUPLED
+  use shr_is_restart_fh_mod, only : init_is_restart_fh, is_restart_fh, is_restart_fh_type
```
jedwards4b, line 18: This module is missing - did you git add it?

jedwards4b (changes requested):
I think that you neglected to add the new module shr_is_restart_fh_mod.

mediator/med_phases_restart_mod.F90
DeniseWorthen: It should be found in our CDEPS share directory (ufs/cdeps_share) which sits at the UFS-WM CDEPS interface level. These are where we have files these files

```
dtypes.h
shr_assert_mod.F90
shr_frz_mod.F90
shr_infnan_mod.F90
```

which are generated via genf90.pl for you (and which we cannot use because of restrictions by CentralOps).

jedwards4b: Looking at directory cdeps/share I see files shr_assert_mod.F90.in, shr_frz_mod.F90.in and shr_infnan_mod.F90.in but no shr_is_restart_fh_mod.F90.in
or shr_is_restart_fh_mod.F90 - that file needs to go in that directory if only to enable github testing of the ufs build.

DeniseWorthen: So a ``shr_is_restart_fh_mod.F90`` inside of cdeps/share will be enough? It doesn't need to be a ".in" file, right?

DeniseWorthen:
I made a PR to CDEPS to add the required module file.

DeniseWorthen:
@jedwards4b I can't tell why this is failing the build check. It seems the line it's complaining about is inside a ifndef CESMCOUPLED.

jedwards4b:
@DeniseWorthen I found the problem - you need to approve https://github.com/ESCOMP/CDEPS/pull/330 so that the new share file is actually built and added to the library.

jedwards4b (approved):

DeniseWorthen:
@jedwards4b  Thanks for the coordination.

PR #552, opened 2025-03-26
Add explicit RH creation for uv3d mapping

### LINKED ISSUE #548: Remove reuse of RHs for the _uv3d mapping options
Currently UFS does not make use of the cart3d mapping in CMEPS. I recently started work on enabling this for UFS and ran into some issues. 

Specifically, mappatch_uv3d and mapbilnr_uv3d currently re-use the RHs for the mappatch and mapbilnr. This means the check in ``med_map_RH_is_created`` returns false for these RHs, since they're not separate RHs. If the RH check is modified to return ``true`` for the uv3d variants, then the mapping of the NormOne fails because it now passes the RH check, whereas previously it did not.

https://github.com/ESCOMP/CMEPS/blob/e203956cb10e8589b3199c563c2f6637607e0328/mediator/med_map_mod.F90#L261-L272

I was able to modify ``med_map_field`` to not fail, but I think ultimately the better and cleaner solution is to allow the _uv3d variant RHs to be created. 

Having separate RHs also allows the switch to using the ``vectorRegrid=true`` option in ESMF 8.6+, which would allow the removal of the user-code in med_map_uv_cart3d in the future, at least for mapping methods which support the vector regridding (UFS needs consf, and that is not currently supported). I tried to get this to work, but I obtained an error I could not resolve (src/Infrastructure/Mesh/src/ESMCI_Mesh_Vector_Regrid.C, line:254:Vector regridding not supported without original coords) so my initial PR for this feature will retain the current cart3d code.

DeniseWorthen:
### Description of changes

Explicitly creates mappatch_uv3d, mapbilnr_uv3d. Adds a mapconsf_uv3d mapping for use by UFS.

### Specific notes

- fixes #548 
- required by https://github.com/ufs-community/ufs-weather-model/issues/2667

This PR does not yet implement the mapping changes required by UFS and so is bfb in current develop branch of UFS. It should be B4B in CESM, but requires testing.

### DIFF
--- a/mediator/med_internalstate_mod.F90
+++ b/mediator/med_internalstate_mod.F90
@@ -77,7 +77,8 @@ module med_internalstate_mod
   integer , public, parameter :: mapfillv_bilnr    = 15 ! fill value followed by bilinear
   integer , public, parameter :: mapbilnr_nstod    = 16 ! bilinear with nstod extrapolation
   integer , public, parameter :: mapconsf_aofrac   = 17 ! conservative with aofrac normalization (ufs only)
-  integer , public, parameter :: nmappers          = 17
+  integer , public, parameter :: mapconsf_uv3d     = 18 ! conservative with uv3d mapping
+  integer , public, parameter :: nmappers          = 18
   character(len=*) , public, parameter :: mapnames(nmappers) = &
        (/'bilnr       ',&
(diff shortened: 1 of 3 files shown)


jedwards4b (approved):
LGTM

Issue #551, opened 2025-03-24
A fix for ROF_NCPL failure in an NCPL consistency test for several compsets


kdraeder:
I don't see any related issues in ESCOMP/CMEPS, so I'm opening one.

I tried to build a case with resolution TL319_t232 and compset G_JRA_RYF in CESM3_0_beta05, but got a rof_cpl_time error:

> ERROR: assume that rof_cpl_time is always greater than or equal to ocn_cpl_time

I didn't change any coupling times in the setup.
OCN_NCPL: 24
ROF_NCPL: 48

I'm guessing the 48 comes from the choice of compset and
components/cmeps/cime_config/config_component_cesm.xml:
```
  <entry id="ROF_NCPL">
    <type>integer</type>
    <default_value>8</default_value>
    <values match="last">
      <value compset="_DATM.*_MOM6.*_DROF"        >[B]$ATM_NCPL[/B]</value>
```

I changed the ATM_NCPL to OCN_NCPL in that xml ROF 'compset' line,
to make it consistent with components/cmeps/cime_config/runseq/runseq_general.py:
 ```
   if rof_cpl_time < ocn_cpl_time:
        expect(False, "assume that rof_cpl_time is always greater than or equal to ocn_cpl_time")
```
The model built successfully, so this looks like a solution, if not the optimal solution.

**Comment by kdraeder** [comment]
@ekluzek @jedwards4b 
If anyone thinks this is the right solution at the moment, I can open a PR about it.
If a more comprehensive solution would be better, then someone familiar with CMEPS should probably handle it.

**Comment by jedwards4b** [comment]
I agree that is the correct solution, @alperaltuntas what do you think?

**Comment by kdraeder** [comment]
Thanks for the guidance!
Is cesm3_0_beta05-1-g8e7888c a good enough version to use in a PR?

**Comment by jedwards4b** [comment]
The best place to start is the head of the development branch for the component you are working on, in this case it would be the 'main' branch of cmeps.

**Comment by kdraeder** [comment]
My cesm clone is in /glade/work/raeder/Models/cesm3_0_beta05.
I synced my github kdraeder fork of cmeps and set my 'origin' to be that fork.
I pulled main from there and checked out a new development branch.
I made my code change and committed it.
Now I'm trying to build the case, but that's failing. 
/glade/derecho/scratch/raeder/Test_ROF_NCPL_bld/bld/csm_share.bldlog.250409-123834:
```
[...]

PR #556, opened 2025-04-10
Make check of ROF_NCPL pass by updating default

kdraeder:
### Description of changes
Components with "_DATM.\*_MOM6.\*_DROF" in their names have ROF_NCPL set to $ATM_NCPL by default, which can fail the check against $OCN_NCPL.  Changed ATM to OCN.

### Specific notes

#551

No

No

### Testing performed
This change enables a resolution TL319_t232 in compset G_JRA_RYF in cesm3.0_alphabranch (recently merged into master) to build.  Tests with cesm3_0_beta05 showed that this change does not prevent running.

### DIFF
--- a/cime_config/config_component_cesm.xml
+++ b/cime_config/config_component_cesm.xml
@@ -379,7 +379,7 @@
     <type>integer</type>
     <default_value>8</default_value>
     <values match="last">
-      <value compset="_DATM.*_MOM6.*_DROF"        >$ATM_NCPL</value>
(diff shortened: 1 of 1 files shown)


billsacks:
@kdraeder - Thanks for your careful investigation but, no, this PR should still come in. The change I made was specific to a particular test configuration that I think is separate from the issue you're fixing here.

jedwards4b (approved):

PR #575, opened 2025-07-07
MEGAN isoprene soil moisture options

fvitt:
### Description of changes
 Implement namelist options for  soil moisture activation factor used in MEGAN isoprene emissions

### Specific notes

### Testing performed

### DIFF
--- a/cesm/nuopc_cap_share/shr_megan_mod.F90
+++ b/cesm/nuopc_cap_share/shr_megan_mod.F90
@@ -33,6 +33,11 @@ module shr_megan_mod
   public :: shr_megan_linkedlist       ! points to linked list of shr_megan_comp_t objects
   public :: shr_megan_mapped_emisfctrs ! switch to use mapped emission factors
   public :: shr_megan_comp_ptr
+  public :: shr_megan_use_gamma_sm
+  public :: shr_megan_min_gamma_sm
+
+  logical :: shr_megan_use_gamma_sm = .true.
+  real(r8) :: shr_megan_min_gamma_sm = 0._r8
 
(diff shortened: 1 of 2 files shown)


wwieder (approved):
Thanks for opening up this PR, @fvitt .

To bring the CTSM side PR to our b4b dev branch, we'll also need this cmeps PR.  Will it still be b4b then, @ekluzek?  

Eventually I think we'll want an answer changing tag that sets `shr_megan_use_gamma_sm = .false.` for CESM3 to get BVOC emissions that are more in line with what the CAM_CHEM group is expecting.  What timeline does this need to come in to a CESM tag?

ekluzek:
@fvitt this adds a place to store the namelist parameters for these. But, I'm assuming the actual effects are only in the CTSM code? As such I'm wondering if the namelist control should just be in CTSM? The main reason I can see for needing it here, would be if the CAM build namelist is going to set these based on specific CAM configurations. If it's only triggered off of CTSM settings (like for instance only what clm6_0 physics is used) -- it should just be in CTSM. But, if both CAM needs to for example change it based on chemistry or CAM physics options, then it needs to be here.

I just want to check that this makes sense.

wwieder:
Thanks for this comment, Erik.  FWIW it was not easy for me to find where this should be set when we were trying to see how Shawn's case we configured.  A simple CLM namelist control would be more obvious (for me), but maybe not consistent with how Megan controls are typically set or modified by users?  I'll defer to the cam-chem folks here.

fvitt:
Long longer needed.  The new gamma_sm options have be moved to CTSM.

PR #597, opened 2025-09-27
Fix gust factors

jedwards4b:
### Description of changes
gust factor changes
### Specific notes

Adam Herrington, Mariana Vertenstein

### Testing performed


adamrher (approved):
shr_flux_mod changes look good. I don't recognize all the namelist_definition changes and so I'm guessing you added those on purpose.

jedwards4b:
@adamrher I think that we will want those, but you're right, they don't belong in this pr and I've removed them.

PR #603, opened 2025-10-20
Add one more barrier at end of run

mnlevy1981:
### Description of changes
In CESM we ran into a funny issue where the ocean component was taking significantly longer to write output at the end of the run than the other components, and somehow the other components convinced the mediator that the model had finished successfully even though the ocean component was still doing stuff. This barrier makes everyone wait to catch up.

### Specific notes

Contributors other than yourself, if any: @jedwards4b wrote the code, @gustavo-marques ran it (I have nothing to do with it but Gustavo is on PTO)

CMEPS Issues Fixed (include github issue #): None

Are changes expected to change answers? this should be bfb unless the baseline run did not write out the last MOM6 history file

### Testing performed
Gustavo reran some cases where his last monthly history files from MOM6 were not being written out because the mediator was exiting successfully before FMS finished; I believe he verified the 47 months of output he got from both runs were bit-for-bit, plus he got the elusive 48th month of MOM6 output. AFAIK nobody has run any CESM tests with this.

### DIFF
--- a/cesm/driver/esm.F90
+++ b/cesm/driver/esm.F90
@@ -9,7 +9,7 @@ module ESM
   use shr_mem_mod  , only : shr_mem_init
   use shr_log_mod  , only : shr_log_setLogunit, shr_log_error
   use esm_utils_mod, only : logunit, maintask, dbug_flag, chkerr
-  use esmf         , only : ESMF_FAILURE
+  use esmf         , only : ESMF_FAILURE, ESMF_VMBARRIER
   implicit none
   private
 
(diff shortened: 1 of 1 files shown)


billsacks (approved):
Thanks, all, for getting this fixed! The fix looks fine to me, but I'm also going to run this by the ESMF team to see if they understand why this is needed. But I think it's fine to merge it if you want to get it in: it's hard to imagine this doing any harm. (Darn, I shouldn't have said that... feels like famous last words!)

PR #622, opened 2025-12-29
Remove a duplicate block from esmFldsExchange_cesm_mod

billsacks:
### Description of changes

This block of code was accidentally duplicated twice.

### Specific notes

Are changes expected to change answers? (specify if bfb, different at roundoff, more substantial) - no: bfb

### Testing performed
From `cesm3_0_alpha08a` ran these two tests:

```
ERS_D_Ld3.ne30pg3_t232.B1850C_LTso.derecho_intel.allactive-defaultio
SMS_D.TL319_t232.G_JRA_RYF.derecho_intel
```

These both passed and are bit-for-bit. (They both had NLCOMP failures, as I've also experienced in other recent testing, but this seems unrelated to the changes here.)


mvertens (approved):
Thanks for finding this.

Issue #624, opened 2026-01-08
Use shr_wv_sat_qsat_liquid for qsat calculation in flux_atmOcn_diurnal_mod


billsacks:
The qsat calculations in flux_atmOcn_Large and flux_atmOcn_COARE_mod were recently updated to use shr_wv_sat_qsat_liquid. flux_atmOcn_diurnal_mod was *not* updated because it is not currently used in production. At some point we should probably make a similar change there.

Note that I'm about to submit a PR that introduces namelist control around this behavior for flux_atmOcn_Large and flux_atmOcn_COARE_mod. flux_atmOcn_diurnal_mod should eventually follow a similar pattern. Until then, the new config variable, aofluxes_use_shr_wv_sat, will only apply to the Large&Pond and COARE schemes.

I see that shr_wv_sat_qsat_liquid also isn't used in flux_atmocn_UA_mod, but it looks like that may be intentional because the UA formulation uses a different qsat formulation.

PR #589, opened 2025-08-29
Shr flux usage

jedwards4b:
### Description of changes
Changes from Jiang-zhu for using shr_flux_mod plus artifacts in testing.

### Specific notes

Contributors other than yourself, if any:

CMEPS Issues Fixed (include github issue #):

Are changes expected to change answers? (specify if bfb, different at roundoff, more substantial) 

Any User Interface Changes (namelist or namelist defaults changes)?

### Testing performed
Please describe the tests along with the target model and machine(s) 
If possible, please also added hashes that were used in the testing

PR #625, opened 2026-01-09
Add namelist control around the qsat calculation for atm-ocn fluxes

billsacks:
### Description of changes

The calculation of qsat in the atm-ocn fluxes was [recently changed](https://github.com/ESCOMP/CMEPS/pull/589) to use the shr_wv_sat module. However, NorESM would like to continue to use the old formulation for now. This PR reintroduces the old code and introduces a new config variable that controls whether to use the old or new formulation for qsat.

The goal here is to allow NorESM and CESM to use the same code, so that @mvertens doesn't need to maintain a separate fork of CMEPS for NorESM. For now, NorESM will need a one-line difference in the namelist definition file, setting the default value of aofluxes_use_shr_wv_sat to false instead of true.

Note that this only applies to ocn_surface_flux_scheme = 0 (Large) or 1 (COARE) (but see #624 regarding extending this at some point).

This PR also renames some arguments in the calls to shr_wv_sat_qsat_liquid to better match the argument names in that subroutine.

### Specific notes

Are changes expected to change answers? (specify if bfb, different at roundoff, more substantial)  No - bfb

Any User Interface Changes (namelist or namelist defaults changes)? Adds new namelist flag, aofluxes_use_shr_wv_sat.

### Testing performed
Ran these tests in the context of cesm3_0_alpha08a (but with share code updated to the latest, since that's a dependency of the latest CMEPS):
```
ERS_D_Ld3.ne30pg3_t232.B1850C_LTso.derecho_intel.allactive-defaultio
SMS_Ld1.ne30_ne30_mg17.FC2010climo.derecho_intel.cam-outfrq1d
ERP_D_Ln9.ne30pg3_ne30pg3_mt232.F1850C_MTso.derecho_intel.cam-outfrq9s
```

I ran 4 versions of these tests:
- ocn_surface_flux_scheme = 0, aofluxes_use_shr_wv_sat = true (out-of-the-box defaults): bit-for-bit
- ocn_surface_flux_scheme = 1, aofluxes_use_shr_wv_sat = true, comparing against baselines with ocn_surface_flux_scheme = 1: bit-for-bit
- ocn_surface_flux_scheme = 0, aofluxes_use_shr_wv_sat = false: changes answers, as expected
- ocn_surface_flux_scheme = 1, aofluxes_use_shr_wv_sat = false, comparing against baselines with ocn_surface_flux_scheme = 1: changes answers as expected

### DIFF
--- a/cesm/flux_atmocn/flux_atmocn_COARE_mod.F90
+++ b/cesm/flux_atmocn/flux_atmocn_COARE_mod.F90
@@ -153,9 +153,12 @@ contains
           endif
 
           if (aofluxes_use_shr_wv_sat) then
+             ! This version uses a qsat calculation method consistent with what's used in CAM
              call shr_wv_sat_qsat_liquid(ts(n), pslv(n), esat_val, qsat_val)
              ssq = 0.98_R8 * qsat_val   ! sea surf hum (kg/kg)
           else
+             ! This version uses the qsat calculation method that was used for many years,
+             ! prior to Aug 2025, and which is still being used by default in NorESM
              ssq = 0.98_R8 * qsat(ts(n)) / rbot(n)   ! sea surf hum (kg/kg)
           end if
 
--- a/cesm/flux_atmocn/flux_atmocn_Large.F90
+++ b/cesm/flux_atmocn/flux_atmocn_Large.F90
@@ -215,9 +215,12 @@ contains
           endif
 
           if (aofluxes_use_shr_wv_sat) then
+             ! This version uses a qsat calculation method consistent with what's used in CAM
              call shr_wv_sat_qsat_liquid(ts(n), pslv(n), esat_val, qsat_val)
              ssq    = 0.98_R8 * qsat_val                ! sea surf hum (kg/kg)
           else
+             ! This version uses the qsat calculation method that was used for many years,
+             ! prior to Aug 2025, and which is still being used by default in NorESM
              ssq    = 0.98_R8 * qsat(ts(n)) / rbot(n)   ! sea surf hum (kg/kg)
           end if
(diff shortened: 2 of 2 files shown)


billsacks:
Thanks, @mvertens . I have made a branch where I've merged the latest escomp changes into the noresm branch. That includes both the recent changes in escomp/main and the changes in this PR. I think that's the first step to getting these reconciled. I'll open a PR into the noresm branch for that. Once we get that in, we should talk about the next steps... I'm not fully sure of it myself, but am happy to talk it through with you.

mediator/med_phases_aofluxes_mod.F90
```diff
@@ -419,6 +420,14 @@ subroutine med_aofluxes_init(gcomp, aoflux_in, aoflux_out, rc)
        add_gusts = .false.
     end if
 
+    call NUOPC_CompAttributeGet(gcomp, name='aofluxes_use_shr_wv_sat', value=cvalue, isPresent=isPresent, isSet=isSet, rc=rc)
+    if (chkerr(rc,__LINE__,u_FILE_u)) return
+    if (isPresent .and. isSet) then
```
jedwards4b: Perhaps we should write to the log whether or not aofluxes_use_shr_wv_sat is set.

mvertens:
@jedwards4b - these are both really good suggestions.

mvertens:
@billsacks - for the NorESM branch you can write that 'This option is currently being used by NorESM'. I'm not sure I would add that further validation needs to be done to adopt the changes for using the share code for now. Does that sound reasonable?

jedwards4b (approved):

PR #634, opened 2026-02-24
For water tracers: Add handling of rank-3 fields and refactor med_phases_prep_glc_map_lnd2glc

billsacks:
### Description of changes

Two sets of changes that will be needed for water tracers - and in particular, for the tracer version of Flgl_qice_elev, which will be the first field in CMEPS that has two ungridded dimensions:

(1) Extend various methods to support rank-3 fields

(2) Refactor med_phases_prep_glc_map_lnd2glc to pre-compute vertical interpolation weights; this will be helpful when we introduce Flgl_qice_elev_wtracers to avoid duplicating this calculation of vertical interpolation weights between the non-tracer (rank-2 on the lnd grid and rank-1 on the glc grid) and tracer (rank-3 on the lnd grid and rank-2 on the glc grid) fields.

### Specific notes

Contributors other than yourself, if any: Claude Code did a lot of the writing of this code, but I have reviewed it all carefully

Are changes expected to change answers? (specify if bfb, different at roundoff, more substantial) bfb when using #633 as a baseline

### Testing performed

In the context of cesm3_0_alpha08e, ran:

(1) aux_glc on derecho plus a few extra tests, with comparisons against baselines generated with #633 .

The extra tests beyond aux_glc were:

```
SMS_Ly2.f09_g17_gris20.T1850Gg.derecho_intel
SMS_Ld5.f10_f10_ais8gris4_mg37.I1850Clm50SpGag.derecho_intel.cism-test_coupling
SMS_Lm13.f10_f10_mg37.I1850Clm50SpG.derecho_intel
ERS_Ld5.ne30pg3_t232.B1850C_LTso.derecho_intel.allactive-decstart
ERS_Ld5.ne30pg3_t232.BHISTC_LTso.derecho_intel.allactive-decstart
```

All tests passed and were bit-for-bit, except for this failure that also failed in the baseline: `FAIL NCK_Ly3.f09_g17_gris20.T1850Gg.derecho_gnu COMPARE_base_multiinst`

(2) Full prealpha testing on derecho and izumi with comparison against cesm3_0_alpha08e

Tests passed except for tests that also failed in the baselines or seemed to have passed due to tweaks made in the baseline (e.g., increasing wallclock time for `SUB_D_Ln9.ne3pg3_ne3pg3_mt232.FHIST.izumi_nag.cam-outfrq9s`).

Baseline failures were all as expected (except that a few B compset tests failed baseline comparison due to a difference in which cpl.hx.ww3 files were present in the run vs. baseline, which presumably is due to something outside of this PR):

```
ERS_Ld5.ne30pg3_t232.B1850C_LTso.derecho_intel.allactive-decstart
ERS_Ld5.ne30pg3_t232.BHISTC_LTso.derecho_intel.allactive-decstart
ERS_Ly7.f09_g17_gris4.T1850Gg.derecho_intel
MULTINOAIS_Ly2.f10_f10_ais8gris4_mg37.I1850Clm50SpRsGag.derecho_intel.cism-change_params
SMS_Lm13.f10_f10_mg37.I1850Clm50SpG.derecho_intel
SMS_D_Ly1.f09_g17_ais8.T1850Ga.derecho_gnu
SMS_D_Ld5_P24x1.f10_f10_ais8gris4_mg37.I1850Clm50SpGag.izumi_nag.cism-test_coupling
```

Note that I showed that these differences go away when comparing against baselines generated with #633 , for similar or identical tests.

So I think it's safe to conclude that this is bit-for-bit with #633 .

### DIFF
--- a/mediator/med_phases_prep_glc_mod.F90
+++ b/mediator/med_phases_prep_glc_mod.F90
@@ -613,7 +613,7 @@ contains
              if (chkerr(rc,__LINE__,u_FILE_u)) return
              if (lndAccum2glc_cnt > 0) then
                 ! If accumulation count is greater than 0, do the averaging
-                data2d(:,:) = data2d(:,:) / real(lndAccum2glc_cnt)
+                data2d(:,:) = data2d(:,:) / real(lndAccum2glc_cnt, R8)
              else
                 ! If accumulation count is 0, then simply set the averaged field bundle values from the land
                 ! to the import field bundle values
@@ -631,7 +631,7 @@ contains
              if (chkerr(rc,__LINE__,u_FILE_u)) return
              if (ocnAccum2glc_cnt > 0) then
                 ! If accumulation count is greater than 0, do the averaging
-                data2d(:,:) = data2d(:,:) / real(ocnAccum2glc_cnt)
+                data2d(:,:) = data2d(:,:) / real(ocnAccum2glc_cnt, R8)
              else
                 ! If accumulation count is 0, then simply set the averaged field bundle values from the ocn
                 ! to the import field bundle values


billsacks:
> The i/o of fields with 2 ungridded dimensions follows the pattern used for fields with 1 ungridded dimension: For fields with 1 ungridded dimension, a separate variable is written for each level (e.g., `lndImp_Flgl_qice_elev1`, `lndImp_Flgl_qice_elev2`, etc.). For fields with 2 ungridded dimensions, a separate variable is written for each combination of levels. So, for example, with 2 water tracers and 11 elevation classes, we'll get 22 variables on the NetCDF file. These are named like `lndImp_Flgl_qice_elev_wtracers1_1`, `lndImp_Flgl_qice_elev_wtracers2_1`, `lndImp_Flgl_qice_elev_wtracers1_2`, `lndImp_Flgl_qice_elev_wtracers2_2`, `lndImp_Flgl_qice_elev_wtracers1_3`, etc., where the indices appear in the same order as the ungridded dimensions.
> 
> Now that I write that out, I'm realizing that there's something kind of confusing here: in this case, the name of the 'elev' dimension comes before the name of the 'wtracers' dimension, but the index of the 'wtracers' dimension comes before the index of the 'elev' dimension. I don't think I want to change the name: various code relies on '_wtracers' being the final suffix for all water tracer fields. Changing the order of the ungridded dimensions might be a bit more doable, though still isn't appealing to me, since @mvertens and I agreed that it made the most sense to put the water tracer dimension first. Maybe the solution is to add some metadata in the NetCDF file giving the meaning of each of these indices. But I'm not sure how easy that will be to do (without resorting to hard-coding stuff in an ugly way), because I'm not sure if we have that information readily available in any data structures that are accessible in med_io_mod. I haven't dug into this yet... maybe it's easier than I'm imagining... @mvertens I'd be happy to talk more with you about this if you think this feels important....

I talked this through with @mvertens . We both agree that it seems best, long-term, to write fields with extra dimensions for the ungridded dimensions, rather than writing a separate field for each permutation of ungridded dimensions. However, the required changes feel like more work than I'm up for taking on right now, so we agreed to leave things as they currently are for now. Changing the i/o code to write fields with one or more extra dimensions probably wouldn't be too hard; somehow storing some metadata on the meaning of each ungridded dimension, so that we can give dimensions meaningful names and share dimensions between fields, would be somewhat more work; but the big time-consuming piece of this would probably be changing the CDEPS cplhist modes to work with files with this new convention.

I am, however, making a change now so that the variable naming on the i/o files is more intuitive: I'm swapping the dimension order between the water tracer dimension and the other ungridded dimension. So now, for Flgl_qice_elev_wtracers, for elevation class 5 and tracer number 2 (for example), the name will be `lndImp_Flgl_qice_elev_wtracers5_2` – so at least the order of integer indices matches the order of suffixes.

billsacks:
A couple of additions / clarifications to my last comment:

(1) Here is a new ESMF issue with a feature request to be able to store and retrieve names of ungridded dimensions: https://github.com/esmf-org/esmf/issues/528. That could be the easiest way to get the needed metadata. Then we could potentially do something like: create a NetCDF dimension for each unique named ungridded dimension; when we encounter another Field with an already-encountered ungridded dimension name, confirm that the bounds match so that we can reuse the already-created NetCDF dimension.

(2) No changes were needed in this PR to accommodate the dimension reordering; I needed to make some changes in some other code that isn't yet in a PR.

PR #637, opened 2026-03-03
Add test components (X comps)

billsacks:
### Description of changes

Add the test components that used to be in CIME. These can be used for testing CMEPS with artificial fields. In CESM, this is done with the `X` compset. These components belong more in CMEPS than in CIME.

This brings in some, but not all history of these X components. I brought in as much history as I could fairly easily without going back to points where the history got very messy (with tons of merge commits that cluttered things up).

Note that, for `X` compsets to continue to work, this should be coordinated with the CIME changes in https://github.com/ESMCI/cime/pull/4946. But it's safe for the CMEPS changes to come in before the CIME changes.

### Specific notes

Are changes expected to change answers? (specify if bfb, different at roundoff, more substantial) no - bfb

### Testing performed

`SMS_Ld2_D_P8x1.f10_g37.X.green_gnu` in cesm3_0_alpha08f with cime and cmeps updated.


fischer-ncar (approved):
Thanks for doing this.

PR #638, opened 2026-03-03
Add water tracers

### LINKED ISSUE #583: Review and update water isotope coupling fields
CMEPS has water isotope coupling fields (search for `wiso`). However, my understanding is that these were just ported over from cpl7 and have never been exercised / tested extensively. For the upcoming water isotope work, we'll need to review and update (if necessary) these water isotope coupling fields.

### LINKED ISSUE #601: Proposed changes for handling water isotopes and water tracers
I'm working on adding a capability to flexibly add water isotopes and water tracers more generally throughout CESM. Our plan is to coordinate this addition through the driver and share code portion of CMEPS. This involves coordinating both the number and names of water tracers present in a simulation and their initial tracer ratio. (Physical parameterizations for isotope fractionation will also be shared between components, but the tentative plan is that that will probably be handled via CESM_share code rather than requiring coordination via the mechanisms discussed here.)

At the Fortran level, my vision (which I have discussed with @mvertens ) is to have a module (in `cesm/nuopc_cap_share`) that stores this information, which both the mediator and other components can use to set up their coupling fields and other data structures needed for water tracers/isotopes. (MOM6 doesn't allow use of share code, so MOM6 - and any other component for which this is an issue - will need a different way to access this information. I'm thinking this can be done by these components directly reading the NUOPC CompAttributes that will store the information coming from the nuopc.runconfig file.)

There are a few key requirements driving our proposed design:
- A user must be able to define any arbitrary number and name of water tracers/tags, and these must be coordinated across all components
- Tracers/tags can be associated with bulk water or with any of the pre-defined isotopes. e.g., we can have a tracer for some specific portion of HDO, where the water is tagged using some specific function/characteristic like other generic water tracers/tag, but which fractionates like standard HDO. Note that this implies that we can have an arbitrary number of tracers for any given water isotope.

The current plan for the user interface, based on discussion with @nusbaume and @jiang-zhu , is the following:
- **The most important point**: We will have a new xml variable defined by CMEPS, WATER_TRACERS, where the user will specify a comma-separated list of tracers to include in this simulation. Non-isotope tracers will just give the name of the tracer; isotope tracers will appear like `TracerName:IsotopeName`, e.g., `MySpecialTracer:HDO` or `AnotherTracer:H218O`. The same formulation will be used for the tracers that represent the full quantity of a given isotope, and we will probably just have a naming convention that their tracer names will match the isotope names. So the full WATER_TRACERS list might look like `MyRegularTracer,HDO:HDO,H218O:H218O,SomeOtherTracer,MyHdoTracer:HDO,YetAnotherTracer` – so there are three tracers/tags of regular water (MyRegularTracer, SomeOtherTracer and YetAnotherTracer), two isotopes (HDO and H218O), and a tracer that is a subset of HDO (MyHdoTracer). (Note that the ordering in the list is unimportant.)
- This XML variable will get parsed by CMEPS's buildnml to create three variables in nuopc.runconfig:
  - water_tracer_names (colon-delimited list of strings; in the above example, this would be `MyRegularTracer:HDO:H218O:SomeOtherTracer,MyHdoTracer:YetAnotherTracer`; this cannot be overridden other than by modifying the WATER_TRACERS xml variable)
  - water_tracer_species (colon-delimited list of strings; in the above example, this would be `:HDO:H218O::HDO:`; note that empty entries correspond to tracers of regular water without a species; this cannot be overridden other than by modifying the WATER_TRACERS xml variable)
  - water_tracer_initial_ratios (probably colon-delimited list of real values; values will default to 1 for every tracer (I can explain the rationale for this default if anyone wants to hear it), but this can be overridden in user_nl_cpl; in the above example, the default will be `1.0,1.0,1.0,1.0,1.0,1.0`). (My original plan was to have dynamically-named config variables like `water_tracer_initial_ratio_MyRegularTracer`, leveraging the CIME changes in https://github.com/ESMCI/cime/pull/4877, but from further discussion with @nusbaume and @jiang-zhu , the feeling was that this isn't important enough to warrant this extra complexity in CIME, and we can just have a list of values.)

These variables in nuopc.runconfig will then get parsed to create an array of data structures containing the name, species name and/or ID, and initial ratio for each tracer.

I'm happy to take any feedback on this plan. I plan to start implementing it in the next couple of days.

billsacks:
### Description of changes

Core changes needed to add water tracers in CESM, for the sake of water isotopes and other purposes.

Major pieces of this include:
- XML and config variables for defining water tracers
- Addition of all necessary water tracers in esmFldsExchange_cesm_mod and fd_cesm.yaml. These use an ungridded dimension for water tracers.
- Calls to a routine to check that test tracers maintain their initial ratios. This ensures that tracers are being treated the same as their bulk counterparts (in terms of mapping, merging, etc.). 
- Addition of water tracer fields in med_test_comps, so this tracer infrastructure can be verified in an X compset test
- Various other changes to some of the med_test_comps to better exercise CMEPS functionality, especially for the lnd -> glc mapping; also added some fields from GLC and to ROF in med_test_comps: Fgrg_rofi and Fgrg_rofl

Some pieces are not yet implemented:
- Including water tracers in the custom lnd->rof mapping
- Including water tracers/isotopes in atm-ocn flux calculations
- Including water tracers in the budget tables (med_diag_mod)

Depends on https://github.com/ESCOMP/CESM_share/pull/82

### Specific notes

- Resolves #583 (though this turned out to be much more extensive than simply reviewing and updating what had been in place before)
- Resolves #601 

Are changes expected to change answers? (specify if bfb, different at roundoff, more substantial) Changes answers for X compsets (due to some changed counter logic and maybe other things); should be bit-for-bit for all other compsets

Any User Interface Changes (namelist or namelist defaults changes)? Adds xml and config variables controlling water tracers

### Testing performed
In the context of cesm3_0_alpha08o, with this branch along with https://github.com/ESCOMP/CESM_share/pull/82, ran CESM prealpha tests on derecho and izumi, and aux_glc tests on derecho, all with baseline comparisons. All failures also failed in the baseline except for `SMS_D_Ln9_P1536x1.ne0CONUSne30x8_ne0CONUSne30x8_mt12.FHIST.derecho_intel.cam-outfrq9s` (which seemed to fail due to https://github.com/ESCOMP/CTSM/issues/3772, which is a sporadic issue), and all baseline comparisons passed. (Note: full testing was done before final commits addressing review points; only minimal additional testing was done after those final commits.)


billsacks:
I still want to do a more extensive review and testing of this myself, so for now marking this as draft.

billsacks:
@nusbaume I have tentatively assigned you as primary reviewer, but I could reassign to someone else.

@mvertens and @Katetc - also looping you in

cime_config/namelist_definition_drv.xml
```diff
@@ -2595,6 +2594,66 @@
     </values>
   </entry>
 
+  <entry id="water_tracer_names" modify_via_xml="WATER_TRACERS" skip_default_entry="true">
+   <type>char</type>
+   <category>water_tracers</category>
+   <group>ALLCOMP_attributes</group>
```
billsacks, line 2588: I'm putting the new variables in ALLCOMP_attributes. I'm not sure if this is right, or if these should go in DRIVER_attributes. I chose ALLCOMP_attributes just because there was more stuff (and more relevant stuff) there, but I'm not sure why I'd choose one over the other.

But actually, maybe ALLCOMP_attributes is needed given that MOM may need to access these rather than going through the share code??

@mvertens do you have guidance here?

mvertens: @billsacks - if any component needs this variable - then it should belong in ALLCOMP_attributes. I think I envisioned the DRIVER_attributes as having variables that are available on all processors but are specific to the driver itself (e.g. pio, etc).
I am not sure why the following variables 
```
     reprosum_diffmax
     reprosum_recompute, 
     reprosum_use_ddpdd
     wv_sat_scheme
     wv_sat_table_spacing
     wv_sat_transition_start
     wv_sat_use_tables
```
are in the DRIVER_attributes and not in the ALLCOMP_attributes. 
@jedwards4b - what are your thoughts on this?

mediator/fd_cesm.yaml
```diff
@@ -186,26 +198,42 @@
        canonical_units: kg m-2 s-1
        description: lnd export to river
      #
-     - standard_name: Flrl_rofdto
```
billsacks, line 189: This removal of `Flrl_rofdto` looks weird in the diff - since it looks like I replaced `Flrl_rofdto` with `Flrl_irrig_wtracers` - but in fact this is intentional: I removed the no-longer-used `Flrl_rofdto` and separately added `Flrl_irrig_wtracers`.

mediator/med_io_mod.F90
```diff
@@ -40,6 +40,8 @@ module med_io_mod
 
   ! private member functions
   private :: med_io_file_exists
+  private :: med_io_def_var_with_atts
+  private :: med_io_read_1d_var
```
billsacks, line 44: The changes in this module are not directly related to the addition of water tracers. Rather, they are a follow-on to the additions made to med_io_mod in https://github.com/ESCOMP/CMEPS/pull/634. Those additions *were* for the sake of adding water tracers, so this is still somewhat related to this overall goal. However, I'd be open to moving these to a separate PR if desired.

mediator/med_phases_prep_glc_mod.F90
```diff
@@ -778,16 +815,10 @@ subroutine med_phases_prep_glc_map_lnd2glc(gcomp, rc)
...
-    ! Map the accumulate land field from the land grid (in multiple elevation classes)
+    ! Map the accumulated land field from the land grid (in multiple elevation classes)
     ! to the glc grid (in multiple elevation classes) using bilinear interpolation
     ! ------------------------------------------------------------------------
 
-    ! Initialize accumulated field bundle on the glc grid to zero before doing the mapping
-    do ns = 1,is_local%wrap%num_icesheets
-       call fldbun_reset(toglc_frlnd(ns)%FBlndAccum2glc_g, value=0.0_r8, rc=rc)
-       if (chkerr(rc,__LINE__,u_FILE_u)) return
-    end do
```
billsacks, line 789: This call to fldbun_reset was duplicated below, so I removed this unnecessary call.

billsacks:
I have run more extensive testing (noted in my edited top-level comment) and given the diffs a final, quick once-over myself. So this is now (finally!) ready for review.

cime_config/testdefs/testmods_dirs/drv/water_tracers/user_nl_cpl
```diff
@@ -0,0 +1,2 @@
+water_tracer_initial_ratios = "2:4"
```
nusbaume: In practice water tracers will almost always be less than one (i.e. less than the total bulk water), so I think it would be good to have one of these tracers (or maybe add an additional set of tracers?) that are fractions of the bulk water instead of multiples.

med_test_comps/xatm/src/atm_comp_nuopc.F90
```diff
@@ -121,6 +123,8 @@ subroutine InitializeAdvertise(gcomp, importState, exportState, clock, rc)
     character(CL)     :: cvalue
     character(len=CL) :: logmsg
     logical           :: isPresent, isSet
+    logical           :: flds_wtracers
```
nusbaume: Nit-pick, but would it help to make this logical variable name a verb like `wtracers_present`?  I always feel like that makes it easier to read in the code below, although it is certainly not critical.

med_test_comps/xshare/dead_nuopc_mod.F90
```diff
@@ -230,6 +275,212 @@ end subroutine SetScalarField
...
+
+    rc = ESMF_SUCCESS
+
+    call shr_wtracers_get_bulk_fieldname( &
+         fieldname = fld%stdname, &
+         is_wtracer_field = is_wtracer_field, &
+         bulk_fieldname = wtracer_bulk_fldname)
+    if (.not. is_wtracer_field) then
+       call ESMF_LogWrite(subname//": ERROR: "//trim(fld%stdname)// &
+            " does not end with the expected suffix for a water tracer field", &
```
nusbaume: Should we include the suffix here as well?


```suggestion
            " does not end with the expected suffix (' "//WTRACERS_SUFFIX//" ') for a water tracer field", &
```

This would require bringing in `WTRACERS_SUFFIX` via a `use` statement.

mediator/med_field_info_mod.F90
```diff
@@ -188,14 +188,12 @@ subroutine med_field_info_array_from_names_wtracers(field_names, field_info_arra
...
     allocate(field_info_array(n_fields))
-    ! For now, hard-code n_tracers, since we haven't set up the tracer information; we'll
-    ! fix this in an upcoming set of changes
-    n_tracers = 0
 
     do i = 1, n_fields
        is_tracer = wtracers_is_wtracer_field(field_names(i))
        if (is_tracer) then
           ! Field is a water tracer; assume a single ungridded dimension
+          n_tracers = wtracers_get_num_tracers()
```
nusbaume, line 202: Would it be better to pull this function outside of the `n_fields` loop?  I realize this might mean adding a `shr_wtracers_present` check first to avoid running into an error.

I am mostly requesting this because it might look like there is a different number of tracers for different fields, which I don't believe can be the case (?).

mediator/med_methods_mod.F90
```diff
@@ -2758,4 +2760,118 @@ subroutine med_methods_check_for_nans_3d(dataptr, nancount)
...
+                     trim(FBName)//":"//trim(fieldNameList(n)))
+             else if (fieldrank == 2) then
+                call ESMF_FieldGet(fieldTracers, farrayPtr=dataTracers3d, rc=rc)
+                if (chkerr(rc,__LINE__,u_FILE_u)) return
+                call ESMF_FieldGet(fieldNonTracer, farrayPtr=dataNonTracer2d, rc=rc)
+                if (chkerr(rc,__LINE__,u_FILE_u)) return
+                call wtracers_check_tracer_ratios(dataTracers3d, dataNonTracer2d, &
+                     trim(FBName)//":"//trim(fieldNameList(n)))
+             else
+                call shr_log_error(subname//": ERROR: unhandled field rank", &
```
nusbaume: Should we write out the value of the bad `fieldrank` either to the error message or to the log?  I feel like that might help save a developer some extra time if they do experience this failure.

mediator/med_methods_mod.F90
```diff
@@ -2758,4 +2760,118 @@ subroutine med_methods_check_for_nans_3d(dataptr, nancount)
...
+                     line=__LINE__, file=u_FILE_u, rc=rc)
+                return
+             end if
+          else
+             ! This is the situation for a small number of fields where we have a tracer
+             ! field without a corresponding non-tracer field.
+             if (dbug_flag > 5) then
+                call ESMF_LogWrite(trim(subname)//": Skipping check for <" // trim(fieldNameList(n)) // &
+                     "> which has no corresponding non-tracer field", ESMF_LOGMSG_INFO)
+             end if
```
nusbaume, line 2875: While we can't compare it to a bulk water field, we can ensure that variables like `roce` (ocean tracer ratios) are maintaining their initial ratio values.

We don't have to do it for this PR, but maybe worth adding in the future at some point?

nusbaume (changes requested):
Thanks for all of this great new tracer infrastructure @billsacks! The only real concern I have is making sure the water tracer test contains at least one initial ratio that is less than one, as that will be closer to production run configurations.  All the other comments are either just questions or totally optional code changes.

cime_config/testdefs/testmods_dirs/drv/water_tracers/user_nl_cpl
billsacks: Good idea; I'll change one of the tracers to be a fraction of bulk water.

med_test_comps/xatm/src/atm_comp_nuopc.F90
billsacks: Good idea - I think I had followed some earlier code here, but I agree that changing this to wtracers_present is more clear. I'll do that (or something similar).

billsacks: Actually, I'll use has_wtracers since wtracers_present is the name of the function so that would be a name collision; also, I noticed that I'm already using has_wtracers in another module so that will be consistent.

mediator/med_field_info_mod.F90
billsacks: Here was my rationale for moving this fetch of n_tracers into the is_tracers conditional (which means putting it inside the loop):

This means we are potentially getting n_tracers multiple times instead of just once, but it allows the code to work in the situation where we don't have any tracer fields and shr_wtracers_mod hasn't been initialized. (I'm not sure if this will arise in practice, but it might.)

I considered pulling this out of the loop and instead putting it in a conditional on whether shr_wtracers_mod has been initialized - otherwise setting n_tracers to 0. But the problem with that is: If this code is called from a path where there *are* water tracers, but shr_wtracers_mod hasn't yet been initialized, the resulting error will be more cryptic.

So I lean towards keeping this code as it is. But if you think it would be helpful, I can add a comment - probably something similar to my above paragraph starting with "This means".

Let me know your thoughts.

cime_config/testdefs/testmods_dirs/drv/water_tracers/user_nl_cpl
billsacks: Fixed in f043f593c

med_test_comps/xatm/src/atm_comp_nuopc.F90
billsacks: Fixed in 35abfa1be

med_test_comps/xshare/dead_nuopc_mod.F90
billsacks: Fixed in 00d88a692

mediator/med_methods_mod.F90
billsacks: Fixed in a847f23cb

mediator/med_methods_mod.F90
billsacks: > While we can't compare it to a bulk water field, we can ensure that variables like `roce` (ocean tracer ratios) are maintaining their initial ratio values.
>
> We don't have to do it for this PR, but maybe worth adding in the future at some point?

Sure, do you want to open an issue for this if it seems worthwhile?

billsacks:
@nusbaume - thank you very much for your thoughtful review! I have addressed most of your suggestions. One outstanding point for discussion is https://github.com/ESCOMP/CMEPS/pull/638#discussion_r3244359550 . Also possibly opening an issue for https://github.com/ESCOMP/CMEPS/pull/638#discussion_r3202722456 ? Beyond that, I'll just work to get the workflow passing... e.g., resolving warnings.

billsacks:
@nusbaume - I have made some other minor changes to get the standalone build to work in the GitHub action. So this is ready for re-review whenever you get a chance... I think the only outstanding things are what I mentioned in my [earlier comment](https://github.com/ESCOMP/CMEPS/pull/638#issuecomment-4455035281).

mediator/med_methods_mod.F90
nusbaume: Great!  Made issue #655 so we can keep track of it going forward.

mediator/med_field_info_mod.F90
nusbaume: Thanks for the explanation @billsacks!  I would much rather be safe than slightly more readable, so I'm happy to leave this line where it is.  Maybe I would vote to add the comment in, though, assuming it doesn't clutter the code or anything?  That way four years from now when I look at this code again I will remember this conversation.

nusbaume (approved):
Thanks for the code changes and responses @billsacks!  Everything looks great to me now.

PR #643, opened 2026-04-01
fix temp arg created

anton-seaice:
### Description of changes

Fixes an error when building with `-check all`

> forrtl: warning (406): fort: (1): In call to MED_IO_WRITE_TIME, an array temporary was created for argument #3

### Testing performed

### DIFF
--- a/mediator/med_phases_restart_mod.F90
+++ b/mediator/med_phases_restart_mod.F90
@@ -348,7 +348,7 @@ contains
              call med_io_enddef(io_file)
           end if
 
-          tbnds = days_since
+          tbnds = (/days_since,days_since/)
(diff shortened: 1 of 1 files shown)


billsacks (approved):
This fix looks good to me - thank you!

Since this is still marked draft, I'm assuming that there's still some more work / testing you want to do? Before we merge, it would be good to see that at least a little testing has been done on this change.

anton-seaice:
Thanks @billsacks 

Indeed - I needed to get back to this and run some tests.

I've confirmed this now builds with `-check all` ok.

I've confirmed the normal Release build has no change in answers - I ran this is a couple of configs, and tests for reproducibility against old results and over restarts. 

In the `cpl.r` file, `time_bnds` appear fine - both the upper and lower bound equal the time variable.

anton-seaice:
Please merge if you are happy !

PR #645, opened 2026-04-14
Remove unused Faxa_ vars

### LINKED ISSUE #644: Some _Faxa_ fields are added but not used in CESM
In this section of the FldsExchange:

https://github.com/ESCOMP/CMEPS/blob/1f8d26a23be9809848146b1334ffa55d1b9d7fa1/mediator/esmFldsExchange_cesm_mod.F90#L1816-L1847

Faxa_sens and Faxa_lat are added from _compatm_ but then nothing is done with them

Both those fluxes are calculated in the _aoflux_ routine instead, and the _aoflux_ calculated fluxes are used.

Is this an error? I think 

       call addfld_from(compatm , 'Faxa_sen')
       call addfld_from(compatm, 'Faxa_lat' )

Can be removed (which means they can also be removed from data forcings)

anton-seaice:
### Description of changes

Removes some unused fields from the CESM fields exchange module, 

### Specific notes

CMEPS Issues Fixed (include github issue #): closes https://github.com/ESCOMP/CMEPS/issues/644

bfb

### Testing performed

Adhoc tests show no impact from this change, results are bfb with previous results.

### DIFF
--- a/mediator/esmFldsExchange_cesm_mod.F90
+++ b/mediator/esmFldsExchange_cesm_mod.F90
@@ -1814,7 +1814,6 @@ contains
     ! to ocn: merged sensible heat flux
     ! ---------------------------------------------------------------------
     if (phase == 'advertise') then
-       call addfld_from(compatm , 'Faxa_sen')
        call addfld_aoflux('Faox_sen')
        call addfld_from(compice , 'Fioi_melth')
        call addfld_to(compocn , 'Foxx_sen')
@@ -1830,7 +1829,6 @@ contains
     ! to ocn: surface latent heat flux and evaporation water flux
     ! ---------------------------------------------------------------------
     if (phase == 'advertise') then
-       call addfld_from(compatm, 'Faxa_lat' )
        call addfld_aoflux( 'Faox_lat' )
        call addfld_aoflux( 'Faox_evap')
        call addfld_to(compocn, 'Foxx_lat' )
--- a/mediator/fd_cesm.yaml
+++ b/mediator/fd_cesm.yaml
@@ -433,16 +433,6 @@
        canonical_units: N m-2
        description: atm import to med - meridional component of momentum flux
      #
-     - standard_name: Faxa_lat
-       alias: mean_laten_heat_flx_atm
-       canonical_units: W m-2
-       description: atm import to med
-     #
-     - standard_name: Faxa_sen
-       alias: mean_sensi_heat_flx_atm
-       canonical_units: W m-2
-       description: atm import to med
-     #
      #-----------------------------------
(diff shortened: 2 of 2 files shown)


billsacks (approved):
Thanks for this cleanup. It looks like these variables are just sent by datm in ERA5 mode, which doesn't appear to be used in CESM. If they *were* used, I think some other changes would be needed in this module. So I agree that removing these is the right thing to do.

I'm just waiting for the workflows to complete their testing, then will merge.

billsacks:
I'm thinking these variables can also be removed from fd_cesm.yaml:

https://github.com/ESCOMP/CMEPS/blob/567e1845f00e3f8047bb6634eec16ca9f0161f9f/mediator/fd_cesm.yaml#L436-L445

@anton-seaice and @mvertens - do you agree?

billsacks:
I see that the failing "scripts regression tests" are unrelated to this PR.

Before merging, I'd like some input/confirmation on the removal of these fields from fd_cesm.yaml (see comment above).

Also, @anton-seaice would you be able to document specific test(s) you ran, so we can have that recorded? If you don't have records of that, I'll run one or more tests so we can have something specific confirming testing that has been done.

uturuncoglu:
@billsacks @anton-seaice please do not remove those fields from the dictionary which could have downstream affect in CDEPS data modes.

anton-seaice:
> Thanks for this cleanup. It looks like these variables are just sent by datm in ERA5 mode, which doesn't appear to be used in CESM. If they _were_ used, I think some other changes would be needed in this module. So I agree that removing these is the right thing to do.

We actually discovered this because we were trying to get the ERA5 datm working.

By removing them here (CMEPS), it means that the ERA5 datm can advertise then but then they don't get connected and so they aren't needed in the ERA5 streams file.


> Also, @anton-seaice would you be able to document specific test(s) you ran, so we can have that recorded? If you don't have records of that, I'll run one or more tests so we can have something specific confirming testing that has been done.

I ran global ACCESS-OM3 at 100km (1deg) and 25km(0.25deg) with the CDEPS JRA forcing for very short tests (6 hours) and confirmed no impact on reproducability with existing output or impacts on determinism (including over restarts). (Links that only mean something to ACCESS people: [1 ](https://github.com/ACCESS-NRI/access-om3-configs/pull/1266) [2](https://github.com/ACCESS-NRI/access-om3-configs/pull/1267) [3](https://github.com/ACCESS-NRI/access-om3-configs/pull/1268))

billsacks:
@uturuncoglu - thanks for clarifying that these are still needed in the field dictionary.

@anton-seaice - thanks for your clarifications, and sorry for introducing confusion on that.

PR #649, opened 2026-04-21
Fix ERS tests for compsets with WW3

alperaltuntas:
### Description of changes

Set `histaux_wav2med_file1_ntperfile` to 1 for ERS and ERI test cases.

### Specific notes

No answer changes

### Testing performed
ERS.TL319_t232_wg37.GW_JRA.derecho_intel

### DIFF
--- a/cime_config/namelist_definition_drv.xml
+++ b/cime_config/namelist_definition_drv.xml
@@ -2182,6 +2182,8 @@
     <desc>Number of time samples per file.</desc>
     <values>
       <value>30</value>
+      <value TESTCASE='ERS'>1</value>
+      <value TESTCASE='ERI'>1</value>
     </values>
   </entry>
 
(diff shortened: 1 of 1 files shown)


billsacks (approved):

billsacks (changes requested):
I just approved... but actually, thinking about it more: can you please add a comment here saying why this is needed? I think I vaguely remember why, but a comment will help when we look back at this in the future.

alperaltuntas:
@billsacks The restart tests were failing because the WW3 aux history files contain multiple time samples, which leads to mismatched filenames and contents between the initial and restart runs. This change writes output to daily files to ensure consistency across initial and restart filenames.

billsacks:
@alperaltuntas thank you for the reminder. I should have been more clear: I was thinking it could be good to add a comment like that in the xml file itself - or a comment in the xml that points to your comment here - so that this somewhat unintuitive setting will be more obvious when we look back at it later.

alperaltuntas:
> @alperaltuntas thank you for the reminder. I should have been more clear: I was thinking it could be good to add a comment like that in the xml file itself - or a comment in the xml that points to your comment here - so that this somewhat unintuitive setting will be more obvious when we look back at it later.

Done.

billsacks (approved):
Thank you!!

Issue #646, opened 2026-04-15
Consider removing scripts_regression_tests workflow, likely replacing it with something else


billsacks:
The srt.yml workflow has been failing recently (due to https://github.com/ESMCI/cime/issues/4965). This raised the question for me as to why we are running scripts_regression_tests from CMEPS. Should we remove this? Perhaps we can replace it with running some small CMEPS-focused test list?

It looks like this was originally added by @fischer-ncar and @jedwards4b .

fischer-ncar:
I believe Jim added this since the CIME scripts_regression_tests depended on CMEPS.  I'm not sure if this is still true or not.

billsacks:
Thanks, @fischer-ncar - that makes sense and it's good to know the background.

My feeling is that we should get rid of running the full scripts_regression_tests and instead run a select set of tests more tailored to exercising cmeps that would hopefully pick up any issues that would impact scripts_regression_tests or any other testing that requires cmeps (which is really all testing in the CESM context).

mvdebolskiy:
Btw, tmate keeps running after failed `srt`, so you have wallclock time of 6h on every commit to an open PR branch, I would suggest to elevate the priority of this issue.

jedwards4b:
I don't understand why you allow this issue to continue to exist.  Either fix the test or comment out the tmate section in the srt.

billsacks:
@jedwards4b - I don't want to pull you too far into this, but since you commented: do you see any strong reason to keep the scripts_regression_tests run in place in CMEPS? I'm tempted to just remove it: I can see running a few system tests as part of the CMEPS GitHub action, but I'm not sure it's worth trying to keep scripts_regression_tests running from within CMEPS.

jedwards4b:
I think that if you remove it you should replace it with some other testing.

billsacks:
Based on discussion with @mvertens , I am temporarily addressing this by disabling the srt workflow - see #652 . I want to leave this issue open for a better long-term solution that will involve adding some new GitHub testing to CMEPS; at that point we can consider completely deleting the srt workflow rather than just disabling it.

PR #652, opened 2026-05-13
Only run srt workflow when manually triggered

billsacks:
### Description of changes

Don't run the CIME scripts_regression_tests workflow on PR and main branch pushes, because it is currently not working.

### Specific notes

CMEPS Issues Fixed (include github issue #): Stop-gap that partially addresses #646.

Are changes expected to change answers? (specify if bfb, different at roundoff, more substantial) No - bfb

### Testing performed
None

### DIFF
--- a/.github/workflows/srt.yml
+++ b/.github/workflows/srt.yml
@@ -2,13 +2,11 @@
 
 name: scripts regression tests
 
-# Controls when the action will run. Triggers the workflow on push or pull request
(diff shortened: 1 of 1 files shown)


billsacks:
@fischer-ncar - sorry for the confusion, here is the correct PR. Diffs are effectively the same as before, but now off of the latest version of cmeps.

fischer-ncar (approved):

Issue #398, opened 2023-08-01
Trying to get mesh out of `cpl_scalars` field


uturuncoglu:
It seems that following if statement is not working properly and tries to get mesh out of `cpl_scalars`.  

https://github.com/ESCOMP/CMEPS/blob/98dcf46c8886104b95cddfd5b02588b3dd9f6722/mediator/med_methods_mod.F90#L390

Just before the if statement the code tries to remove the `cpl_scalars` from the field list but it is still in the FB and if it is the first field then CMEPS fails with following error (I put couple of more print statement to be sure) because it uses `cpl_scalars`.

```
20230731 224553.040 INFO             PET00 (med_methods_FB_init):FB FBImpatm_wav fieldNameList from STflds
20230731 224553.040 INFO             PET00 (med_methods_FB_init):Sa_u10m
20230731 224553.040 INFO             PET00 (med_methods_FB_init):Sa_v10m
20230731 224553.040 INFO             PET00 (med_methods_FB_init):cpl_scalars
20230731 224553.040 INFO             PET00 (med_methods_FB_init):cpl_scalars removed ...
20230731 224553.040 INFO             PET00 (med_methods_FB_init):Sa_u10mloop again
20230731 224553.040 INFO             PET00 (med_methods_FB_init):Sa_v10mloop again
20230731 224553.040 INFO             PET00 (med_methods_FB_init):cpl_scalars mesh from STgeom
20230731 224553.040 ERROR            PET00 ESMF_GeomBase.F90:997 ESMF_GeomBaseGet Value unrecognized or out of range  -  Grid not geometry type
20230731 224553.040 ERROR            PET00 ESMF_FieldGet.F90:487 ESMF_FieldGetDefault Value unrecognized or out of range  - Internal subroutine call returned Error
20230731 224553.040 ERROR            PET00 med_methods_mod.F90:431 Value unrecognized or out of range  - Passing error in return code
20230731 224553.040 ERROR            PET00 med.F90:1744 Value unrecognized or out of range  - Passing error in return code
20230731 224553.040 ERROR            PET00 MED:src/addon/NUOPC/src/NUOPC_ModelBase.F90:1639 Value unrecognized or out of range  - Passing error in return code
[...]

Issue #311, opened 2022-10-03
Can't run with HIST_OPTION=nstep?


mnlevy1981:
I am trying to run for a couple of days with coupler history file every time step to see what fields coming into MOM6 look like. I ran

```
./xmlchange HIST_N=1,HIST_OPTION=nstep
```

and now I'm `PET####.ESMF_LogFile` output with messages like

```
$ cat PET0000.ESMF_LogFile
20221003 101948.374 ERROR            PET0000 (med_time_alarmInit): unknown option nstep
20221003 101948.374 ERROR            PET0000 ESMF_Alarm.F90:1445 ESMF_AlarmSet() Object being used before creation  - Bad Object
20221003 101948.374 ERROR            PET0000 med_phases_history_mod.F90:224 Object being used before creation  - Passing error in return code
20221003 101948.374 ERROR            PET0000 MED:src/addon/NUOPC/src/NUOPC_ModelBase.F90:2215 Object being used before creation  - Passing error in return code
20221003 101948.374 ERROR            PET0000 /glade/p/cesmdata/cseg/PROGS/build/28560/esmf-8.2.0b23/src/Infrastructure/Trace/src/ESMCI_Trace.C:1512 ESMCI:TraceEventPhaseExit() Wrong argument specified  - Trace regions not properly nested exiting from region: [MED] med_phases_history_write Expected exit from: MED:(med_phases_history_write)
20221003 101948.374 ERROR            PET0000 /glade/p/cesmdata/cseg/PROGS/build/28560/esmf-8.2.0b23/src/Infrastructure/Trace/src/ESMCI_Trace.C:1470 ESMCI::TraceEventCompPhaseExit() Wrong argument specified  - Internal subroutine call returned Error
20221003 101948.374 ERROR            PET0000 ESMCI_FTable.C:832 ESMCI_FTableCallEntryPointVMHop Wrong argument specified  - Internal subroutine call returned Error
20221003 101948.374 ERROR            PET0000 ESMCI_FTable.C:1100 c_esmc_compwait Wrong argument specified  - Internal subroutine call returned Error
20221003 101948.374 ERROR            PET0000 ESMF_Comp.F90:1255 ESMF_CompExecute Wrong argument specified  - Internal subroutine call returned Error
20221003 101948.374 ERROR            PET0000 ESMF_GridComp.F90:1905 ESMF_GridCompRun Wrong argument specified  - Internal subroutine call returned Error
[...]

PR #285, opened 2022-04-20
use mesh file instead of grid name

jedwards4b:
### Description of changes
Use comparisons of Mesh file instead of grid to determine if components are on the same grid.
This was done so that vertical component used in grid name does not affect this test.  

### Specific notes

Contributors other than yourself, if any: mvertens

CMEPS Issues Fixed: #284 

Are changes expected to change answers? (specify if bfb, different at roundoff, more substantial) 

Any User Interface Changes (namelist or namelist defaults changes)?

### Testing performed

Testing performed if application target is CESM:
- [ ] (recommended) CIME_DRIVER=nuopc scripts_regression_tests.py
   - machines:
   - details (e.g. failed tests):
- [ ] (recommended) CESM testlist_drv.xml
   - machines and compilers:
   - details (e.g. failed tests):
- [ ] (optional) CESM prealpha test
   - machines and compilers
   - details (e.g. failed tests):
- [ ] (other) please described in detail
   - machines and compilers
   - details (e.g. failed tests):

Testing performed if application target is UFS-coupled:
- [ ] (recommended) UFS-coupled testing
   - description:
   - details (e.g. failed tests):

Testing performed if application target is UFS-HAFS:
- [ ] (recommended) UFS-HAFS testing
   - description:
   - details (e.g. failed tests):

### Hashes used for testing:

- [ ] CESM:
  - repository to check out: https://github.com/ESCOMP/CESM.git
  - branch/hash:
- [ ] UFS-coupled, then umbrella repostiory to check out and associated hash:
  - repository to check out:
  - branch/hash:
- [ ] UFS-HAFS, then umbrella repostiory to check out and associated hash:
  - repository to check out:
  - branch/hash:

PR #657, opened 2026-05-26
Dart lowres cmeps

kdraeder:
This issue stems from cime #4933, which is about developing a large ensemble test
motivated by DART applications.

Because of the large ensemble, the testing will be more managable
if it uses a coarse resolution grid.  An ne3 grid is available for CAM and CTSM,
and now a ~10 degree resolution is available in MOM6 (MOM_interface #311).
These have been combined into a new CESM grid and used in ERI and MCC tests,
which also use a new testmod tailored to DART needs.
I'm open to suggestions for a shorter testmod name,
but @billsacks and I feel that it will be helpful to have DART in it.
This grid (especially the MOM6 grid) limits the tasks/instance to 12 
(6 for MOM, 6 for the other components).

An MCC test for a small ensemble passes all test stages
(/glade/work/raeder/Exp/CESM+DART_testing/MCC_cG.ne3pg3_10deg.B_DART.lowres)
but ensembles which require more than 1 node mostly fail
with an error in cmeps/cesm/driver/ensemble_driver.F90.
This seems to arise from smaller ensembles fitting into a single (develop qeueu) node,
where the exact number of processors needed is assigned to them,
while larger ensembles need multiple (cpu/main) nodes 
and more processors are assigned to the job than are requested.
For example, 40 instances request 12 x 40 = 480 processors.
This requires 4 nodes x 128 = 512 processors are assigned.
This difference causes an error:
PetCount (  512) - Async IOtasks (  0) must be evenly divisable by number of members ( 40).
   
When the check for this error is removed, the job goes farther, 
but hangs just before the time stepping in CAM.  This can be prevented by choosing MAX_TASKS_PER_NODE in a way that prevents any instance from being laid out across 2 nodes.
The changes required to do this are beyond the scope of this PR, 
and are handled in [CESM #398](https://github.com/ESCOMP/CESM/pull/398/changes).
 
### Description of changes
Commenting out the consistency check between PetCount and number\_of\_members,
`if(modulo(PetCount-pio_asyncio_ntasks*number_of_members, number_of_members) .ne. 0) then`
allows the test to proceed.
I could not trace the variables back through ESMF to figure out an if-test
which would handle this situation, and developers I talked to weren't certain that it's essential,
so my temporary solution is to comment out the test, without removing it.

### Specific notes

@billsacks @jedwards4b 

CMEPS Issues Fixed (include github issue #): [#641](https://github.com/ESCOMP/CMEPS/issues/641)
This is also essential for issues in other components:
[ESMCI/cime #4933](https://github.com/ESMCI/cime/issues/4933) (overview issue)
[CESM PR #398 ](https://github.com/ESCOMP/CESM/pull/398)
[ESMCI/ccs_config PR #285](https://github.com/ESMCI/ccs_config_cesm/pull/285)
[NCAR/MOM6 #413](https://github.com/NCAR/MOM6/issues/413)

This is not expected to change answers in tests which ran successfully before this change.
Some tests which would not run before will now run.   It's possible that some of those should not run,
but I have not looked into those.

Users who want to run ERI or MCC tests with an ensemble which can fit some, 
but not all, instances on 1 node, will need to include the test_mods developed in  [CESM #398](https://github.com/ESCOMP/CESM/pull/398/changes)
and follow the instructions for setting MAX_TASKS_PER_NODE.

### Testing performed
If possible, please also added hashes that were used in the testing.
Extensive testing (development) of ERI and MCC tests were conducted in a version of cesm3_0_alpha08d,
modified to enable the 10-degree MOM6 grid, using a BHIST compset, on derecho.
The relevant changes (multiple components) were imported to the cesm3_0_alpha09a tag 
and tested in cases in /glade/work/raeder/Exp/CESM+DART_testing:

1. 40 instance MCC; alpha09a_MCC_cG_C40.ne3pg3_10deg.B_DART.MAX_TASKS120
2. 2 instance ERI; alpha09a_ERI_cG_C2_Ld8.ne3pg3_10deg.B_DART.aux_lowres

### DIFF
--- a/cesm/driver/ensemble_driver.F90
+++ b/cesm/driver/ensemble_driver.F90
@@ -254,12 +254,13 @@ contains
     if (chkerr(rc,__LINE__,u_FILE_u)) return
 
     ntasks_per_member = PetCount/number_of_members - pio_asyncio_ntasks
-    if(modulo(PetCount-pio_asyncio_ntasks*number_of_members, number_of_members) .ne. 0) then
-       write (msgstr,'(a,i5,a,i3,a,i3,a)') &
-            "PetCount (",PetCount,") - Async IOtasks (",pio_asyncio_ntasks*number_of_members,") must be evenly divisable by number of members (",number_of_members,")"
-       call ESMF_LogSetError(ESMF_RC_ARG_BAD, msg=msgstr, line=__LINE__, file=__FILE__, rcToReturn=rc)
-       return
-    endif
+!     if(modulo(PetCount-pio_asyncio_ntasks*number_of_members, number_of_members) .ne. 0) then
+!        write (msgstr,'(a,i5,a,i3,a,i3,a)') &
+!             "PetCount (",PetCount,") - Async IOtasks (",pio_asyncio_ntasks*number_of_members, &
+!             ") must be evenly divisable by number of members (",number_of_members,")"
+!        call ESMF_LogSetError(ESMF_RC_ARG_BAD, msg=msgstr, line=__LINE__, file=__FILE__, rcToReturn=rc)
+!        return
+!     endif
 
     !-------------------------------------------
     ! Loop over number of ensemblel members
@@ -280,7 +281,7 @@ contains
        asyncio_task = .false.
        ! Determine pet list for driver instance
        if(pio_asyncio_ntasks > 0) then
-          do n=pio_asyncio_rootpe,pio_asyncio_rootpe+pio_asyncio_stride*(pio_asyncio_ntasks-1),pio_asyncio_stride
+          do n= pio_asyncio_rootpe, pio_asyncio_rootpe+pio_asyncio_stride*(pio_asyncio_ntasks-1), pio_asyncio_stride
              asyncio_petlist(iopetcnt) = (inst-1)*(ntasks_per_member+pio_asyncio_ntasks) + n
              if(asyncio_petlist(iopetcnt) == localPet) asyncio_task = .true.
              iopetcnt = iopetcnt+1
--- a/cesm/driver/ensemble_driver.F90
+++ b/cesm/driver/ensemble_driver.F90
@@ -281,7 +281,7 @@ contains
        asyncio_task = .false.
        ! Determine pet list for driver instance
        if(pio_asyncio_ntasks > 0) then
-          do n= pio_asyncio_rootpe, pio_asyncio_rootpe+pio_asyncio_stride*(pio_asyncio_ntasks-1), pio_asyncio_stride
+          do n=pio_asyncio_rootpe,pio_asyncio_rootpe+pio_asyncio_stride*(pio_asyncio_ntasks-1),pio_asyncio_stride
              asyncio_petlist(iopetcnt) = (inst-1)*(ntasks_per_member+pio_asyncio_ntasks) + n
              if(asyncio_petlist(iopetcnt) == localPet) asyncio_task = .true.
              iopetcnt = iopetcnt+1
--- a/cesm/driver/ensemble_driver.F90
+++ b/cesm/driver/ensemble_driver.F90
@@ -254,13 +254,18 @@ contains
     if (chkerr(rc,__LINE__,u_FILE_u)) return
 
     ntasks_per_member = PetCount/number_of_members - pio_asyncio_ntasks
-!     if(modulo(PetCount-pio_asyncio_ntasks*number_of_members, number_of_members) .ne. 0) then
-!        write (msgstr,'(a,i5,a,i3,a,i3,a)') &
-!             "PetCount (",PetCount,") - Async IOtasks (",pio_asyncio_ntasks*number_of_members, &
-!             ") must be evenly divisable by number of members (",number_of_members,")"
-!        call ESMF_LogSetError(ESMF_RC_ARG_BAD, msg=msgstr, line=__LINE__, file=__FILE__, rcToReturn=rc)
-!        return
-!     endif
+    if(modulo(PetCount-pio_asyncio_ntasks*number_of_members, number_of_members) .ne. 0) then
+       write (msgstr,'(a,i5,a,i3,a,i3,a)') &
+            "PetCount (",PetCount,") - Async IOtasks (",pio_asyncio_ntasks*number_of_members, &
+            ") must be evenly divisable by number of members (",number_of_members, ")."  
+       call ESMF_LogWrite(trim(subname)//": "//trim(msgstr), ESMF_LOGMSG_INFO)
+       write (msgstr,'(3a)') &
+            "Multi-instance (multi-driver): set NINST, NTASKS and MAX_TASKS_PER_NODE ", &
+            "so that the instances are evenly divided across the implied number of nodes ", &
+            "and no extra tasks are requested."
+       call ESMF_LogSetError(ESMF_RC_ARG_BAD, msg=msgstr, line=__LINE__, file=__FILE__, rcToReturn=rc)
+       return
+    endif
 
     !-------------------------------------------
     ! Loop over number of ensemblel members


billsacks:
@kdraeder thank you for your continued work on this.

I had to get my head back into this. I've read back through some of the discussion from a couple of months ago. I have a couple of questions / requests:

(1) @kdraeder - My impression is that we never determined whether this error check was actually needed or not; is that your impression, too?

(2) @kdraeder - It sounds like only select tests pass with this error check removed - that you needed to make some careful changes to PE layouts for the tests to pass. Can you confirm: for these passing tests with carefully selected PE layouts, is the removal of this error check still actually needed? (I couldn't tell this for sure from reading through your comments.)

(3) @kdraeder - If we go ahead with this, please go ahead and actually remove these lines of code rather than commenting them out.

(4) @mvertens @DeniseWorthen @theurich @uturuncoglu @danrosen25 @oehmke - can any of you see a reason why the error check removed in this PR might actually be needed? My understanding of the error check is that it is effectively ensuring that we only use full nodes, but I'm not sure why that would be a requirement... though I may also be misinterpreting the intent of the error check.

kdraeder:
> @kdraeder thank you for your continued work on this.

And thank you for coming back to it.
> 
> (1) @kdraeder - My impression is that we never determined whether this error check was actually needed or not; is that your impression, too?

Yes.  I can see some value in having a correct, robust test here, but I don't know how to implement it, and it's conceivable that the required information is not available there.

> 
> (2) @kdraeder - It sounds like only select tests pass with this error check removed - that you needed to make some careful changes to PE layouts for the tests to pass. Can you confirm: for these passing tests with carefully selected PE layouts, is the removal of this error check still actually needed? (I couldn't tell this for sure from reading through your comments.)

As far as I know, all tests that passed before will pass with the check removed.
Those are tests in which all instances fit on 1 node, or each instance gets its own node(s).
In between those are tests which can fit several, but not all, instances on a node,
which (mostly) cannot pass the check.  Removing it and using the default PE layout 
results in those intermediate jobs hanging.   
Carefully specifying the PE layout enables those tests to succeed.  
> 
> (4) @mvertens @DeniseWorthen @theurich @uturuncoglu @danrosen25 @oehmke - can any of you see a reason why the error check removed in this PR might actually be needed? My understanding of the error check is that it is effectively ensuring that we only use full nodes, but I'm not sure why that would be a requirement... though I may also be misinterpreting the intent of the error check.

The check does issue a somewhat helpful error message instead of allowing the program to proceed to a point where it hangs with no helpful message.  It might be sufficient if the message said something like

> MAX_TASKS_PER_NODE must be set to 
> PEs_per_instance_assigned_by_ESMF * num_instances_which_fit_on_a_node.
> The first can be extracted from the PET files in $rundir.   
> The second is INT[(NINST * requested_tasks_per_inst) / machine_PEs_per_node]

danrosen25:
The code looks right to me
@DeniseWorthen wrote the change in June 2023. It previously validated that the number of ensemble members could be evenly distributed to the PETs. It looks like it was updated so that `pio_asyncio_ntasks` could be given to each member, meaning the `PetCount` should be  `((ntasks_per_member + pio_asyncio_ntasks) * number_of_members)`. If you're runtime configuration doesn't account for adding `pio_asyncio_ntasks` to **each member** then there are probably reasons for it to fail.

https://github.com/ESCOMP/CMEPS/commit/cec8db8d09fa0a0b016d197a68edc67cbd100d97#diff-510899343ac8719363762a83df95863cb3c3219c3be7b48562110f7f1e06398dL202-R262

What sounds like is happening is that you're configuring the application to provided all `512` processors, and this is not divisible by the number of members, `40`. Try running your application via `mpirun -np 480` and telling your batch selection to only allocate 120 mpi tasks per node for better balancing `#PBS -l select=4:ncpus=120:mpiprocs=120` or `#PBS -l select=4:ncpus=128:mpiprocs=120`

DeniseWorthen:
@danrosen25 I can't imagine I would have written the change? It is in cesm/driver/ensemble_driver.F90, which is unused in UFS.

kdraeder:
@danrosen25 
thanks for your perspective.  I think that does represent what's happening under the covers.
I'm approaching it from the perspective of a user who sets a liimited number of "high level"
parameters, in a testing scenario, and then hopes that they translate into a good layout.
In my case I was limited to asking for 12 tasks/instance, but was given 17 PEs. 
I'm not aware of formula for figuring that out ahead of time, or of an xml build variable 
that will let me request N instances/node, so I was hoping that the testing scripts 
would figure out a workable distribution.  I think that you're proposing changes
which would make that happen.

danrosen25:
@kdraeder It sounds more like a CIME/CESM workflow or runtime configuration issue. In the runtime configuration file, does is specify the the total number of PETs or is this something automatically determined by the available PEs? I believe the system provided 12.8 PETs (I'm not sure if there is ESMF managed threading providing other types of PEs so I assume everything is an MPI task) per ensemble member, which isn't allowed. This would signal that the rest of the code, including modeling components, may not support a mix of 12 or 13 PETs depending on their instance.

I'm not a CIME/CESM engineer but I worked on CMEPS, mediator work, six years ago for a different project. So I'm not familiar with the application workflow.

kdraeder:
@danrosen25 I agree that the root cause, and maybe solution, to this issue may be in the CIME/CESM workflow, but I also think that this test in CMEPS makes assumptions about layout that are not robust.

In the test setup we can only specify NINST, NTASKS (tasks per instance per component), and ROOTPE (the layout of each instance).  Threading is inactive by default.  In my test
   NTASKS = 6 for each component
   ROOTPE = 0 for each, except MOM is assigned 6   (so 12 tasks per instance are requested)
   NINST = any reasonable number; we've been using 40 as an example.
From these, TOTAL_PES is calculated to be 40 * 2 * 6 = 480, which can fit on 4 nodes.
So COST_PES is 4 * 128 = 512.  That's also how PetCount is calculated (by ESMF_VMGet), 
which is different from how you outlined that it should be calculated.  
If it were calculated as you outlined (to yield 480), then the check would not fail,
but I don't know what side effects would pop up.  

But even if PetCount lets the check pass, there's still the problem of some instance(s) being layed out on 2 nodes.  The current code does not prevent this, so the user is responsible for anticipating it, which can be challenging.
The xml variable for controlling this is MAX_TASKS_PER_NODE, which seems to be used
in the `mpirun -np` and `mpiprocs=` job definition.
The test would be more useful if it compared MAX_TASKS_PER_NODE (or the local equivalent)
with number_inst_per_node * PES_per_instance.  PES_per_instance is not necessarily 
the number of tasks per instance requested by the user.  An example is 15 instances:
`15*12 = 180` tasks are requested, so 2 nodes are assigned.  Then `2*128/15 = 17` PEs are assigned
to each instance.  Instance 8 is laid out on both nodes, which causes the job to hang.

billsacks:
@kdraeder - thank you for sharing all of these additional thoughts. And @danrosen25 - thank you for sharing your thoughts here as well.

I'm still not clear on the answer to the question I was trying to ask (maybe not clearly) in (2): For the tests where you have carefully specified the PE layout (so that they pass with this error check commented out), do those specific tests pass if this error check is in place, or do even those tests with carefully specified PE layouts fail this error check?

I haven't totally wrapped my mind around the discussion about instances crossing a node, but to me that feels like a different issue. It isn't clear to me why there should be a limitation preventing an instance from crossing a node: this clearly works for a single instance.

As both @kdraeder and @danrosen25 are suggesting, the real issue here may be more related to how CIME sets up the processor layout and batch job submission for multi-instance cases. But I haven't absorbed everything from the comments here or the related issues / PRs, so I could be missing or misunderstanding something.

My key conclusion here is: **It feels to me that there are a lot of subtleties and interacting issues here, and this will require a significant time investment to solve robustly. I am reluctant to remove this error check without a more comprehensive understanding of what's going on and what the right solutions are.** Unfortunately, we (myself and other CESM software engineers) are overloaded right now, and so it may be a while before we have the necessary time to really dig into this. I'd like to understand if you have a workaround for now that will let you move forward even with this error check in place, or if some kind of bypass of this error check is needed. Given that most cases hang when removing this error check, I'm reluctant to just remove it, but if this is a critical blocker for you, we could do something like introduce a for-experts-only namelist variable that bypasses this error check, which you could set in your testing.

kdraeder:
@billsacks thanks for suggesting that I round out the testing.
A test **with the PetCount check active** and 14 instances divided evenly onto 2 nodes
with MAX_TASKS_PER_NODE = 7 * 17 = 119 succeeded.

A more challenging test with 15 instances (8 on the first node and 7 on the second)
and MAX_TASKS_PER_NODE = 96 (8 inst * 12 tasks_per_inst), fails in the PetCount check:
`Invalid argument  - PetCount (  192) - Async IOtasks (  0) must be evenly divisable by number of members ( 15)`
So PetCount was calculated from num_nodes * MAX_TASKS_PER_NODE, 
but that is more than is usable by the number of instances.

It seems highly likely to me that distributing 15 instances evenly onto 3 nodes,
with MAX_TASKS_PER_NODE = (15/3) * 12 = 60  would work (but be inefficient).

**So it appears that the check can stay in, but a more helpful error message might have this added:**
"If this is a multi-instance (multi-driver) case, NINST, NTASKS and MAX_TASKS_PER_NODE 
must be chosen so that the instances are evenly distributed across the implied number of nodes 
and no extra tasks are requested."

It would be helpful to add something like this to the instructions in env_mach_pes.xml
andor the documentation about setting up cases, to help users avoid defining cases which won't run.
If there's agreement about that, I'll work on it.

I agree with @billsacks that making CIME (and ESMF) set up these more challenging cases correctly
could take a lot of work and discussion, and can be delayed.

kdraeder:
@billsacks @danrosen25 do either of you object to my less-intrusive solution?

danrosen25:
PetCount is usually calculated as the number of PEs that were given to the application during execution such as mpirun -np <mpi tasks "PEs">. The workflow is probably calculating the number of mpi tasks based on MAX_TASKS_PER_NODE and blindly passing in the value to the mpirun command.

billsacks:
@kdraeder - my understanding from reading back through this is that your proposal at this point is to return to having this error check in place, but to improve the error message (and possibly other documentation) to be more helpful. If so, I support that plan - and thank you!

I still haven't gotten my head completely around the details here, but your suggested message in https://github.com/ESCOMP/CMEPS/pull/657#issuecomment-4616594853 sounds correct to me. I also like the idea of adding this to the CIME documentation (I think that is what you were suggesting); I see we have a [documentation page on multi-instance](https://esmci.github.io/cime/versions/master/html/ccs/examples/multi-instance.html), and that could be updated to include this information and anything else you feel is relevant.

kdraeder:
@billsacks That's my status.  I've written a slightly different message,
but it's too long, so I split it into a message via ESMF_LogWrite
and one via ESMF_...Error...   But the LogWrite doesn't appear anywhere (including PET files).
Let me know if that sounds familiar, but don't dig into it.

oehmke:
Just checking in about your issue with ESMF_LogWrite(), are you writing to the default log and what do you have logmsgflag set to? 

Also, how long is the error message that you’re trying to write with LogError? 

Thanks!

kdraeder:
@oehmke Thanks for checking in.  I believe that 
`components/cmeps/cesm/driver/ensemble_driver.F90: SetModelServices`
should write to the default log, because the calls to LogWrite don't have a file or unit argument:
`call ESMF_LogWrite(trim(subname)//": "//trim(msgstr), ESMF_LOGMSG_INFO)`
I don't know the file name of the default log.  

The existing message, which I think should be kept, is
 "PetCount (",PetCount,") - Async IOtasks (",pio_asyncio_ntasks*number_of_members, &
 ") must be evenly divisable by number of members (",number_of_members, ")." 

I want to add 
"Multi-instance (multi-driver): set NINST, NTASKS and MAX_TASKS_PER_NODE ", &
"so that the instances are evenly divided across the implied number of nodes ", &
"and no extra tasks are requested."
but that makes the (message + context-added-by-LogWrite) too long (> 512 characters ?).
So I thought I'd put the original message into a LogWrite, which would allow the program
to proceed to the LogSetError, which would have the added message,
which would appear after the original.

kdraeder:
@oehmke Following up on @billsacks suggestion (email) that I
./xmlchange CREATE_ESMF_PET_FILES=TRUE
I see that the LogWrite message appears in both the PET log files and the cesm.log file.
The PET log files were being created when the program died, so I didn't think 
that I needed to activate them.

I also see that the LogWrite message appeared in the cesm.log file when CREATE... was FALSE.
Somehow my searching didn't see that.  Apologies!

So my message works if users look in the cesm.log file as well as the PET log files.
The output is in 
/glade/derecho/scratch/raeder/MCC_cG_C15.ne3pg3_10deg.B_DART.2x96+PRINTs/MCC_cG_C15.ne3pg3_ne3pg3_t232.BHISTC_LT_DART.derecho_intel.allactive-DART_BHIST_lowres--mom-tx10deg.20260612_171421_h1e5fw/run

billsacks:
@kdraeder - great, glad you got this working. What you described sounds reasonable - i.e., that the more extended error message appears in the cesm.log always, and in the PET files if you set CREATE_ESMF_PET_FILES=TRUE. Let me know once you have the changes ready for review.

kdraeder:
@billsacks I tested this with layout parameters which should succeed 
and then with some that should fail and print the new messages.  
I got the expected results, so I'm satisfied with this code.

billsacks (approved):
Thank you for clarifying this error message and for all of your work iterating on this!

PR #659, opened 2026-06-02
Sync w/ latest UFS changes

DeniseWorthen:
### Description of changes

Adds recent changes for CMEPS in UFS

### Specific notes

Adds ufs_trace wrapper to enable profiling for use in UFS. 
Adds capability to use mapfiles in UFS global applications.
Removes HYCOM field exchanges in UFS/HAFS mode (HYCOM has been removed from our component list).
Removes deprecated flux scheme for UFS (ocean_flux_scheme == -1)
Updates ufs/ccpp as required for ccpp changes

ufs_tracing capability was developed by @DusanJovic-NOAA 

Changes should not touch any CESM code and should be B4B.


billsacks (approved):
This looks good to me. I have reviewed the modules that apply to CESM; I haven't looked at modules that apply only to UFS.

@DeniseWorthen do you feel this warrants a review from anyone else?

@fischer-ncar - although this looks safe to me, it's worth keeping an extra eye on the nightly baseline tests after this is merged to verify that this passes all tests and is bit-for-bit.

DeniseWorthen:
@billsacks Thanks for reviewing. In this case, I don't think needs additional review.

PR #662, opened 2026-06-05
fix two unitialized variables found w/ valgrind

### LINKED ISSUE #661: uninitialized variables found by valgrind
Several uninitialized variables were found by @DusanJovic-NOAA using valgrind on a UFS configuration:

==124287== Conditional jump or move depends on uninitialised value(s)
==124287==    at 0xB174239: _gfortran_string_len_trim (in /usr/lib64/libgfortran.so.5.0.0)
==124287==    by 0xB17428B: _gfortran_string_trim (in /usr/lib64/libgfortran.so.5.0.0)
==124287==    by 0x6F7DE9A: __esmflds_MOD_med_fldlist_findname (esmFlds.F90:195)
and

==124287== Conditional jump or move depends on uninitialised value(s)
==124287==    at 0xAEB3425: bcmp (vg_replace_strmem.c:1237)
==124287==    by 0xB17406A: _gfortran_compare_string (in /usr/lib64/libgfortran.so.5.0.0)
==124287==    by 0x6F43EB4: __med_merge_mod_MOD_med_merge_auto_multi_fldbuns (med_merge_mod.F90:131)

The first of these can be fixed with

```
diff --git a/mediator/esmFlds.F90 b/mediator/esmFlds.F90
index 7a2959f..d7522dd 100644
--- a/mediator/esmFlds.F90
+++ b/mediator/esmFlds.F90
@@ -45,8 +45,8 @@ module esmflds
   !-----------------------------------------------

   type, public :: med_fldList_entry_type
-     character(CS) :: stdname
-     character(CS) :: shortname
+     character(CS) :: stdname = ''
+     character(CS) :: shortname = ''

      ! Mapping fldsFr data - for mediator import fields
      integer      , allocatable :: mapindex(:)
```

and the second (most likely a typo)

```
diff --git a/mediator/med_merge_mod.F90 b/mediator/med_merge_mod.F90
index 6d12fa92..887ab23a 100644
--- a/mediator/med_merge_mod.F90
+++ b/mediator/med_merge_mod.F90
@@ -128,7 +128,7 @@ contains
              call med_fld_GetFldInfo(fldptr, compsrc=compsrc, merge_fields=merge_fields, merge_type=merge_type, merge_fracname=merge_fracname, rc=rc)

-             if (merge_type /= 'unset' .and. merge_field /= 'unset') then
+             if (merge_type /= 'unset' .and. merge_fields /= 'unset') then
                    ! If merge_field is a colon delimited string then cycle through every field - otherwise by default nm
                    ! will only equal 1
                 num_merge_colon_fields = merge_listGetNum(merge_fields)
```

DeniseWorthen:
### Description of changes

Initialize two character strings and correct typo.

Contributors other than yourself, if any: @DusanJovic-NOAA 

- fixes #661 

Changes are B4B in UFS testing.

### DIFF
--- a/mediator/esmFlds.F90
+++ b/mediator/esmFlds.F90
@@ -45,8 +45,8 @@ module esmflds
   !-----------------------------------------------
 
   type, public :: med_fldList_entry_type
-     character(CS) :: stdname
-     character(CS) :: shortname
+     character(CS) :: stdname = ''
+     character(CS) :: shortname = ''
 
      ! Mapping fldsFr data - for mediator import fields
      integer      , allocatable :: mapindex(:)
--- a/mediator/med_merge_mod.F90
+++ b/mediator/med_merge_mod.F90
@@ -17,7 +17,7 @@ module med_merge_mod
   use esmFlds               , only : med_fldList_findName
   use perf_mod              , only : t_startf, t_stopf
   use shr_log_mod           , only : shr_log_error
-  
+
   implicit none
   private
 
@@ -128,7 +128,7 @@ contains
              call med_fld_GetFldInfo(fldptr, compsrc=compsrc, merge_fields=merge_fields, merge_type=merge_type, merge_fracname=merge_fracname, rc=rc)
              if (chkerr(rc,__LINE__,u_FILE_u)) return
 
-             if (merge_type /= 'unset' .and. merge_field /= 'unset') then
+             if (merge_type /= 'unset' .and. merge_fields /= 'unset') then
                    ! If merge_field is a colon delimited string then cycle through every field - otherwise by default nm
                    ! will only equal 1
                 num_merge_colon_fields = merge_listGetNum(merge_fields)
@@ -489,7 +489,7 @@ contains
        write(errmsg,*) trim(subname),' input field ungriddedUbound ',ungriddedUbound_in(1),&
             ' for '//trim(merge_fldname), &
             ' not equal to output field ungriddedUbound ',ungriddedUbound_out,' for '//trim(fldname_out)
-      
+
        call shr_log_error(errmsg, rc=rc)
    endif
 
@@ -647,7 +647,7 @@ contains
           endif
           if (wgtfound) then
              if (lbound(dataPtr,1) /= lbound(wgt,1) .or. ubound(dataPtr,1) /= ubound(wgt,1)) then
-                
+
                 call shr_log_error(trim(subname)//": ERROR wgt wrong size", &
                      line=__LINE__, file=u_FILE_u, rc=dbrc)
                 return


mediator/esmFlds.F90
```diff
@@ -45,8 +45,8 @@ module esmflds
   !-----------------------------------------------
 
   type, public :: med_fldList_entry_type
-     character(CS) :: stdname
-     character(CS) :: shortname
+     character(CS) :: stdname = ''
+     character(CS) :: shortname = ''
```
billsacks, line 49: I'm fine with this change, but it makes me wonder if this might be masking a more serious issue. I don't have experience with valgrind, so I'm not sure: does the issue picked up in #661 suggest an actual attempt to reference a field with previously-uninitialized values, or just that valgrind can't guarantee that the data was initialized? If the former, I'm wondering if the real solution is to figure out the code path that has led to the use of a field with uninitialized stdname / shortname? Otherwise, it seems like we might be masking the problem by satisfying valgrind but with non-sensical data. @DeniseWorthen @DusanJovic-NOAA what do you think?

mediator/med_merge_mod.F90
```diff
@@ -128,7 +128,7 @@ subroutine med_merge_auto_multi_fldbuns(coupling_active, FBOut, FBfrac, FBImp, f
              call med_fld_GetFldInfo(fldptr, compsrc=compsrc, merge_fields=merge_fields, merge_type=merge_type, merge_fracname=merge_fracname, rc=rc)
              if (chkerr(rc,__LINE__,u_FILE_u)) return
 
-             if (merge_type /= 'unset' .and. merge_field /= 'unset') then
+             if (merge_type /= 'unset' .and. merge_fields /= 'unset') then
```
billsacks, line 131: I agree with this change - thanks.

billsacks (commented):
Thanks a lot for these changes. I'm interested in hearing your thoughts on my comment below.

mediator/esmFlds.F90
DeniseWorthen: My initial fix was more complicated (partly a result of querying AI for assistance). Those code changes were
```
diff --git a/mediator/esmFlds.F90 b/mediator/esmFlds.F90
index 7a2959f4..18b832b0 100644
--- a/mediator/esmFlds.F90
+++ b/mediator/esmFlds.F90
@@ -188,11 +188,17 @@ contains
     lastfld => fields
     found = .false.
     do while(associated(lastfld%next))
-       if (trim(stdname) == trim(lastfld%stdname)) exit
+       if (trim(stdname) == trim(lastfld%stdname)) then
+          found = .true.
+          exit
+       end if
        lastfld => lastfld%next
     enddo
+
     ! Check the lastfld
+    if (.not. found) then
        if (trim(stdname) == trim(lastfld%stdname)) found = .true.
+    end if
```
The attempt here was to ensure that if a match was not found, lastfld%next would still point to the very last (found) field. But Dusan reported that it did not resolve the issue. 

I also don't have valgrind experience, but Dusan provided instructions and I can try to work through a better fix if you have ideas.

billsacks: With Claude's help, I think I understand the problem: When `med_fldList_findName` is called on a currently-empty list, the last line (`if (trim(stdname) == trim(lastfld%stdname)) found = .true.`) will be working on uninitialized data.

Now that I understand that issue, the fix here makes sense to me.

billsacks (approved):

PR #668, opened 2026-07-01
Small adjustment to the testlist removing strange PE layout for a test

### LINKED ISSUE #666: 32 task tests don't make sense...
In 203526609 there were two tests added that use only a fraction of a node, on either Izumi or Derecho with 32 tasks. This doesn't really make sense when Derecho nodes have 128 processors and Izumi 48. I don't gather anything from the tests or commits that a partial node was the intent. Hence, it seems like these should be increased to use full nodes.

ekluzek:
### Description of changes

Slightly change the testlist. Remove some tests that made no sense.

### Specific notes

  Fixes #666

Are changes expected to change answers? (specify if bfb, different at roundoff, more substantial) b4b

### Testing performed
 Ran aux_cmeps out of cesm3_0_alpha09d making sure for working tests it compares to the baseline I ran
cesm3_0_alpha09d
cmeps hash: 3b1fba9cb55ad26f864d380cc3fd689f5cbf6830
cdeps with cdeps1.0.99

Tests ran as expected

### DIFF
--- a/cime_config/testdefs/testlist_drv.xml
+++ b/cime_config/testdefs/testlist_drv.xml
@@ -195,22 +195,38 @@
       <option name="wallclock"> 00:20:00 </option>
     </options>
   </test>
-  <test compset="I1850Clm50SpG" grid="f10_f10_mg37" name="ERS_Lm13_P32">
+  <test compset="I1850Clm50SpG" grid="f10_f10_mg37" name="ERS_Lm13">
     <machines>
       <machine name="derecho" compiler="nvhpc" category="aux_cmeps"/>
       <machine name="izumi" compiler="intel" category="aux_cmeps"/>
     </machines>
     <options>
       <option name="wallclock"> 00:20:00 </option>
+      <option name="comment">Multi month restart test with CLM-SP and active GLC model CISM"</option>
     </options>
   </test>
-  <test compset="I2000Clm50BgcCropRtm" grid="f10_f10_mg37" name="SMS_D_Ld5_P32" testmods="rtm/default">
+  <test compset="I2000Clm50BgcCropRtm" grid="f10_f10_mg37" name="SMS_D_Ld5" testmods="rtm/default">
     <machines>
       <machine name="derecho" compiler="gnu" category="aux_cmeps"/>
+      <machine name="derecho" compiler="intel" category="aux_cmeps"/>
+      <machine name="derecho" compiler="nvhpc" category="aux_cmeps"/>
       <machine name="izumi" compiler="nag" category="aux_cmeps"/>
     </machines>
     <options>
       <option name="wallclock"> 00:20:00 </option>
+      <option name="comment">DEBUG full CLM BGC-Crop case using RTM with extra RTM output, most compilers"</option>
+    </options>
+  </test>
+  <test compset="I2000Clm60FatesSpRsGs" grid="f45_f45_mg37" name="SMS" testmods="clm-FatesColdSatPhen">
+    <machines>
+      <machine name="derecho" compiler="gnu" category="aux_cmeps"/>
+      <machine name="derecho" compiler="intel" category="aux_cmeps"/>
+      <machine name="derecho" compiler="nvhpc" category="aux_cmeps"/>
+      <machine name="izumi" compiler="nag" category="aux_cmeps"/>
+    </machines>
+    <options>
+      <option name="wallclock"> 00:20:00 </option>
+      <option name="comment">Test of FATES-SP without DEBUG on for most compilers</option>
     </options>
   </test>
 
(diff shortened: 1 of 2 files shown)


billsacks (commented):
I appreciate the fix for the nvhpc issue (it still seems likely to be a compiler issue to me, but I appreciate your submitting this workaround).

Can you give more justification for the test additions? I'm especially wondering about the addition of Fates tests. The aux_cmeps test list needs a major overhaul so I don't yet have clear ideas about what it should and shouldn't include, but my general philosophy would be to keep it fairly small to facilitate quick/cheap turnaround, and only add tests that give additional code coverage of the CMEPS code. Does the Fates test cover CMEPS code that isn't already covered by the existing I compset tests?

billsacks:
@ekluzek - based on our discussion earlier about how aux_cmeps needs a bigger, long-term overhaul, how would you feel about backing out at least the addition of the Fates tests. I'm on the fence about the other test changes you made: the removal of _P32 makes sense, at least.

ekluzek:
> @ekluzek - based on our discussion earlier about how aux_cmeps needs a bigger, long-term overhaul, how would you feel about backing out at least the addition of the Fates tests. I'm on the fence about the other test changes you made: the removal of _P32 makes sense, at least.

I'm am amiable to removing the Fates test if you'd like me to. If the aux_cmeps testlist isn't really viable, I'm not sure it matters much. But, I'll take it out if you'd prefer.

ekluzek:
> I appreciate the fix for the nvhpc issue (it still seems likely to be a compiler issue to me, but I appreciate your submitting this workaround).
> 

I had assumed the problem was that chkErr was used in other subroutines besides the one. But, you are right they are all confined to one subroutine, so this change is not necessary. Hence, it must be a compiler issue for the nvhpc compiler. I figure it's not a bad change to make anyway, because having a basic utility like chkErr usable throughout the module is not a bad thing.

ekluzek:
I've done enough testing that I think this is ready to merge in. So @billsacks let me know if I should delete the Fates test, and then we can merge it in as the next tag.

billsacks:
I think that, unless the Fates tests actually add significant code coverage of CMEPS (e.g., a metric is: are there some changes in CMEPS that would clearly be caught by Fates tests but not by non-Fates tests, e.g., because the coupling fields or other coupling logic are different in Fates configurations?) I'd like them to be backed out. Even though I plan to overhaul the aux_cmeps test list at some point, it will still be confusing to have extra tests that aren't really needed - since I'll probably use the existing tests as a rough guide for what's needed in a new test list - so I'd prefer to only have new tests added there if they add important test coverage of cmeps that's missing with existing tests.

ekluzek:
> I think that, unless the Fates tests actually add significant code coverage of CMEPS (e.g., a metric is: are there some changes in CMEPS that would clearly be caught by Fates tests but not by non-Fates tests, e.g., because the coupling fields or other coupling logic are different in Fates configurations?) I'd like them to be backed out. Even though I plan to overhaul the aux_cmeps test list at some point, it will still be confusing to have extra tests that aren't really needed - since I'll probably use the existing tests as a rough guide for what's needed in a new test list - so I'd prefer to only have new tests added there if they add important test coverage of cmeps that's missing with existing tests.

Perfect. I like that as the criteria, do FATES tests change anything in how CMEPS operates? FATES will be more common, but it seems doubtful that it changes anything in how CMEPS operates or the fields passed to CMEPS. The only thing that comes to mind is that currently: megan, drydep, and fire-emission fields can't be passed for FATES, but that is something that will be fixed. For CMEPS it would probably be good to make sure there are tests that run with all of those and others that don't. The case for having them all off would already be covered by tests without CLM. So the thing to make sure is there, would be atest with all of them on.

billsacks:
Thanks, @ekluzek . Once FATES is the standard configuration it will make sense to switch tests to use that, but in the meantime I think it makes sense for aux_cmeps and other test lists in non-CTSM components to use the more standard CTSM configuration to avoid an explosion of system tests.

Once you back out those test additions I'll merge this.

billsacks:
I just merged #686 , which has an identical fix to the core issue fixed in this PR. I realized that this PR was never merged because of outstanding discussion related to the tests here. @ekluzek I'll leave it to you whether you want to update this PR - bringing in the now-merged changes from #686 and adjusting testing as discussed above - or simply close this PR at this point.

ekluzek:
> I just merged #686 , which has an identical fix to the core issue fixed in this PR. I realized that this PR was never merged because of outstanding discussion related to the tests here. @ekluzek I'll leave it to you whether you want to update this PR - bringing in the now-merged changes from #686 and adjusting testing as discussed above - or simply close this PR at this point.

Hey @billsacks I did the update. So there's not much there now, but it removes some nonsense tests in the list. So seems worth taking in.

If that sounds good, go ahead and merge it, I don't think I need new testing for it. But, I can do that if you'd like.

billsacks (approved):

PR #677, opened 2026-07-14
added new coupling mode for noresm

mvertens:
### Description of changes
Added new valid value to noresm to the COUPLING_MODE xml variable

### Specific notes
This change permits namelist values for coupling_mode to now include noresm. Conditionals throughout the code now can test for both noresm and cesm. 
This xml variable can also now be included in CDEPS config variables to distinguish between noresm and cesm settings.
Not that this PR also will need to be included for upcoming WW3 PRs from noresm to cesm.

CMEPS Issues Fixed:

### Testing performed
`ERS_Ld5.ne30pg3_t232.B1850C_LTso.derecho_intel.allactive-defaultio` from cesm3_0_alpha09d - passes and bit-for-bit. I ran this test both with (a) just the first commit on this branch, and (b) the 3 commits that included Erik's suggested extra parentheses.

### DIFF
--- a/mediator/med_phases_prep_ice_mod.F90
+++ b/mediator/med_phases_prep_ice_mod.F90
@@ -90,7 +90,8 @@ contains
     if (chkerr(rc,__LINE__,u_FILE_u)) return
 
     ! Apply precipitation factor from ocean (that scales atm rain and snow to ice) if appropriate
-    if (trim(coupling_mode) == 'cesm' .and. is_local%wrap%flds_scalar_index_precip_factor /= 0) then
+    if ( (trim(coupling_mode) == 'cesm' .or. trim(coupling_mode) == 'noresm') .and. &
+         is_local%wrap%flds_scalar_index_precip_factor /= 0) then
 
        ! Note that in med_internal_mod.F90 all is_local%wrap%flds_scalar_index_precip_factor
        ! is initialized to 0.
--- a/mediator/med_phases_prep_ocn_mod.F90
+++ b/mediator/med_phases_prep_ocn_mod.F90
@@ -509,7 +509,7 @@ contains
        end do
        ! Compute sw export to ocean bands if required
        if (export_swnet_by_bands) then
-          if (trim(coupling_mode) == 'cesm') then
+          if (trim(coupling_mode) == 'cesm' .or. trim(coupling_mode) == 'noresm') then
              c1 = 0.285; c2 = 0.285; c3 = 0.215; c4 = 0.215
              Foxx_swnet_vdr(:) = c1 * Foxx_swnet(:)
              Foxx_swnet_vdf(:) = c2 * Foxx_swnet(:)
@@ -630,7 +630,8 @@ contains
     end if
 
     ! Apply precipitation factor from ocean (that scales atm rain and snow back to ocn ) if appropriate
-    if (trim(coupling_mode) == 'cesm' .and. is_local%wrap%flds_scalar_index_precip_factor /= 0) then
+    if ((trim(coupling_mode) == 'cesm' .or. trim(coupling_mode) == 'noresm') .and. &
+         is_local%wrap%flds_scalar_index_precip_factor /= 0) then
 
        ! Note that in med_internal_mod.F90 all is_local%wrap%flds_scalar_index_precip_factor
        ! is initialized to 0.
--- a/cime_config/buildnml
+++ b/cime_config/buildnml
@@ -134,9 +134,16 @@ def _create_drv_namelists(case, infile, confdir, nmlgen, files):
     elif case.get_value("RUN_TYPE") == "branch":
         config["run_type"] = "branch"
 
-    if config['COMP_WAV'] == 'ww3' and config['COMP_ICE'] == 'cice':
-        config["wav_ice_coupling"] = "on"
+    # determine coupling mode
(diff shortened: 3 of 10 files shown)


billsacks (approved):
This looks good to me, though I'm curious how you're going to set COUPLING_MODE correctly for noresm and how, if at all, this will interact with CIME_MODEL.

mvertens:
This is one thing that needs to be different for cesm and noresm. The default for CMEPS in noresmhub will be noresm and the default for cmeps in ESCOMP will be cesm. So at this point it does not interact with CIME_MODEL - which is a much bigger lift. For noresm - this will still be CIME_MODEL will still be cesm. Hopefully with the CIME refactor it will be easier to pull that out.

mediator/med_phases_prep_ice_mod.F90
```diff
@@ -90,7 +90,8 @@ subroutine med_phases_prep_ice(gcomp, rc)
     if (chkerr(rc,__LINE__,u_FILE_u)) return
 
     ! Apply precipitation factor from ocean (that scales atm rain and snow to ice) if appropriate
-    if (trim(coupling_mode) == 'cesm' .and. is_local%wrap%flds_scalar_index_precip_factor /= 0) then
+    if ( (trim(coupling_mode) == 'cesm' .or. trim(coupling_mode) == 'noresm') .and. &
```
ekluzek, line 93: I think this might need parenthesis around the top level connected by the or, so that it's clear that the and doesn't only get connected to the "== noresm" part.

ekluzek:
```suggestion
    if ( ( (trim(coupling_mode) == 'cesm' .or. trim(coupling_mode) == 'noresm')) .and. &
```

mediator/med_phases_prep_ocn_mod.F90
```diff
@@ -630,7 +630,8 @@ subroutine med_phases_prep_ocn_custom(gcomp, rc)
     end if
 
     ! Apply precipitation factor from ocean (that scales atm rain and snow back to ocn ) if appropriate
-    if (trim(coupling_mode) == 'cesm' .and. is_local%wrap%flds_scalar_index_precip_factor /= 0) then
+    if ((trim(coupling_mode) == 'cesm' .or. trim(coupling_mode) == 'noresm') .and. &
```
ekluzek: This one might need the same addition of parenthesis as above.

ekluzek:
```suggestion
    if (((trim(coupling_mode) == 'cesm' .or. trim(coupling_mode) == 'noresm')) .and. &
```

ekluzek (approved):
This makes sense to me as well. It all looks correct to me. I didn't see anything of concern in skimming over the code, other than I'm wondering about if some of the logic statements need paranthesis (even if just for clarity).

billsacks:
Unless my eyes are playing tricks on me, I think the last two commits were unnecessary: I think these `.or.` expressions were already parenthesized and now they are doubly-parenthesized. That's not really a problem, just unnecessarily parenthetical.

@mvertens as far as I'm concerned, go ahead and merge this when you're ready.

billsacks:
@mvertens - oh, I remembered I was going to send you a test you could run to confirm that this is bit-for-bit for CESM.

I'd suggest ERS_Ld5.ne30pg3_t232.B1850C_LTso.derecho_intel.allactive-defaultio

You can run this from cesm3_0_alpha09d with this CMEPS update, and compare against the baselines here:

/glade/campaign/cesm/cesmdata/cesm_baselines

(That's the default baseline root for CESM.)

billsacks:
@mvertens - Actually, I'll just go ahead and run that test for you while I'm thinking of it.

ekluzek:
> Unless my eyes are playing tricks on me, I think the last two commits were unnecessary: I think these `.or.` expressions were already parenthesized and now they are doubly-parenthesized. That's not really a problem, just unnecessarily parenthetical.
> 
> @mvertens as far as I'm concerned, go ahead and merge this when you're ready.

Crap you are right. I wasn't looking at it right when I made the suggestion. Sorry about that.

billsacks:
I have run `ERS_Ld5.ne30pg3_t232.B1850C_LTso.derecho_intel.allactive-defaultio` from cesm3_0_alpha09d - this passes and is bit-for-bit. I ran this test both with (a) just the first commit on this branch (c86256ed), and (b) the 3 commits that included Erik's suggested extra parentheses (c86256ed, b5a2ccf6, 13c30fa4).

@mvertens I'll let you decide when you're ready to merge this, including whether you want to keep the latest commits or revert to your original version before the extra parentheses (if the latter, I'm personally okay with you force pushing to this PR to remove the last two commits from history... and would actually slightly prefer that as a less error-prone approach - something like `git checkout c86256ed; git push --force mvertens feature/add_noresm_coupling_mode_to_escomp`)

mvertens:
@billsacks - thank you for running the test! I have done the force push you suggested - and have backed out the last two commits. This should be ready to be merged.

mvertens:
Do you want to merge this or should I? I'm fine with either.

billsacks:
@mvertens - sorry: I think I gave you the wrong commands: the `git checkout` should have been `git reset --hard c86256ed3`. I just did the right commands and force pushed... merging now.

Issue #675, opened 2026-07-07
Rework irrigation coupling for the sake of water tracers


billsacks:
There is special-purpose mapping of irrigation here:

https://github.com/ESCOMP/CMEPS/blob/806cce1de93ac96dcdc977e30f18c1206e8b6bf6/mediator/med_phases_prep_rof_mod.F90#L441-L680

This currently doesn't handle water tracer fields.

I started thinking about how we want to do the volr-based normalization for tracer fields – such as whether each tracer should normalize by its own tracer's volr or by the bulk volr. But then I realized that we probably should do a more substantial rework of irrigation coupling to ensure the correct handling of tracers - see https://github.com/ESCOMP/CTSM/issues/512#issuecomment-4898296928.

When we address that CTSM issue, we'll also need to rework the CMEPS irrigation coupling, both to handle water tracers and to handle any other changes to the LND-ROF irrigation coupling.

billsacks:
Once this is resolved, we should remove the bypassing of the water tracer check for the irrigation field (though that field may no longer exist at this point): see the user_nl_cpl change in billsacks/CMEPS@98cefc20e

PR #678, opened 2026-07-14
Allow skipping checks of some tracer fields

billsacks:
### Description of changes

The irrigation tracer field is not yet mapped correctly (see https://github.com/ESCOMP/CMEPS/issues/675). In order to allow both irrigation and volr to be spatially-varying but to still have tracer checks for everything else, we now exclude the irrigation field from these tracer checks.

**This depends on the CESM_share changes in https://github.com/ESCOMP/CESM_share/pull/85**

### Specific notes

Are changes expected to change answers? (specify if bfb, different at roundoff, more substantial) - no: bfb

Any User Interface Changes (namelist or namelist defaults changes)? Adds water_tracers_variables_not_checked namelist item

### Testing performed

`SMS_D_Ld2.f19_g17.X.green_gnu.drv-water_tracers--drv-glc_avg_frequently` in cesm3_0_alpha09d, with share updated to share1.1.21

### DIFF
--- a/mediator/med_methods_mod.F90
+++ b/mediator/med_methods_mod.F90
@@ -2795,6 +2795,7 @@ contains
     integer :: fieldrank
     logical :: hasSuffix
     logical :: isPresentNonTracer
+    logical :: check_skipped
     type(ESMF_Field) :: fieldTracers
     type(ESMF_Field) :: fieldNonTracer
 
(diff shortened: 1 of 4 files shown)


mediator/med_methods_mod.F90
```diff
@@ -2831,9 +2832,16 @@ subroutine med_methods_FB_check_wtracers(FB, rc)
...
           call ESMF_FieldBundleGet(FB, fieldName=fieldNameNonTracer, isPresent=isPresentNonTracer, rc=rc)
           if (chkerr(rc,__LINE__,u_FILE_u)) return
-          if (isPresentNonTracer) then
+          if (.not. isPresentNonTracer) then
+             ! This is the situation for a small number of fields where we have a tracer
+             ! field without a corresponding non-tracer field.
+             if (dbug_flag > 5) then
+                call ESMF_LogWrite(trim(subname)//": Skipping check for <" // trim(FBName)//"%"//trim(fieldNameList(n)) // &
+                     "> which has no corresponding non-tracer field", ESMF_LOGMSG_INFO)
+             end if
```
billsacks, line 2841: This looks like a bigger diff than it really is: The conditional has been rearranged so that this `.not. isPresentNonTracer` is the first block, and then the standard case is in the `else`. This made sense at first because there were going to be multiple skip conditions so it was going to look like:

```Fortran
if (.not. isPresentNonTracer) then
   ...
else if (variable_in_skip_list) then
   ...
else
   ! standard case, not skipped
end if
```

It ended up making more sense to put the variable-in-skip-list check inside the main code branch - because it was simplest to check/set this in the wtracers_check_tracer_ratios call - but I felt like I marginally liked this conditional block arrangement better because it lends itself better to possible future changes (such as new ways to skip a variable), and I had already done testing with this code structure, so I decided to leave this change in place.

nusbaume (approved):
Thanks @billsacks!

PR #681, opened 2026-07-17
Make sure there are default values for some fields

billsacks:
### Description of changes

This is critical for histaux_wav2med_file1_flds (in case we're not running with ww3); for aofluxes_use_shr_wv_sat, this default isn't currently needed, but adding it is helpful for some future-proofing.

### Specific notes

Contributors other than yourself, if any: @mvertens 

Are changes expected to change answers? (specify if bfb, different at roundoff, more substantial) bfb

Any User Interface Changes (namelist or namelist defaults changes)? Sets a default for histaux_wav2med_file1_flds when not running with ww3 (to an empty list)

### Testing performed
aux_cime_baselines from cesm3_0_alpha09d (`SMS_Ld2.ne30pg3_t232.B1850C_LTso.derecho_intel.allactive-defaultio` hasn't finished; all others pass and are bit-for-bit)

### DIFF
--- a/cime_config/namelist_definition_drv.xml
+++ b/cime_config/namelist_definition_drv.xml
@@ -957,8 +957,8 @@
       (Currently only relevant for ocn_surface_flux_scheme = 0 or 1.)
     </desc>
     <values>
+      <value>.true.</value>
       <value coupling_mode="noresm">.false.</value>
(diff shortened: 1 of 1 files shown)


mvertens (approved):
Thanks for the fixes.

fischer-ncar (approved):

PR #683, opened 2026-07-22
Add WW3 daily-averaged wave fields needed for the NorESM merge

alperaltuntas:
### Description of changes

This PR is needed to sync WW3 forks of CESM and NorESM. The NorESM WW3 cap exports 11 new Sw_*_avg fields and expects the mediator to output them in a daily-averaged auxiliary history file. This defines those fields in fd_cesm.yaml, and drops the stale ww3 histaux overrides that were still requesting the old instantaneous fields.

### Specific notes

This PR must be evaluated and merged in conjunction with:
 - https://github.com/ESCOMP/WW3/pull/42
 - https://github.com/ESCOMP/WW3_interface/pull/19

No answer changes

The only user interface changes is the default value and description of histaux_wav2med_file1_doavg.

### Testing performed
aux_ww3

PR #690, opened 2026-08-13
merge noresm cmeps1.1.57_noresm_v0

mvertens:
### Description of changes
Merge changes from noresm cmeps1.1.57_noresm_v0 back to ESCOMP

### Specific notes
In particular, this has enthalpy computation changes in the mediator that would occur for noresm if CAM computed the enthalpy. However, although these hooks are in CMEPS - the computation of enthalpy in CAM has not been brought over to the main development branch yet.

Also - this PR might potentially interfere with PR https://github.com/ESCOMP/CMEPS/pull/683
So it's not clear which should potentially go first.

CMEPS Issues Fixed: None

Are changes expected to change answers?  bfb

### Testing performed
Testing  has not been done yet.


cime_config/buildnml
```diff
@@ -53,9 +55,14 @@ def _create_drv_namelists(case, infile, confdir, nmlgen, files):
     for val in ("HIST", "REST", "STOP"):
         config[val.lower()+"_option"] = case.get_value(val+"_OPTION")
 
-
     config["comp_ocn"] = case.get_value("COMP_OCN")
 
+    config["set_histaux_l2x1yrg"] = (
+        ".true." if case.get_value("COMP_OCN") == 'blom' and case.get_value("COMP_LND") == 'clm' else ".false."
+    )
```
billsacks: This turning on of histaux_l2x1yrg depending on the ocean and land model seems unusual and not at all intuitive. If it's needed, I'm okay with it, but can you add a comment explaining why it's needed?

cime_config/buildnml
```diff
@@ -122,11 +129,26 @@ def _create_drv_namelists(case, infile, confdir, nmlgen, files):
...
+            )
+            expect(
+                case.get_value("DRV_RESTART_POINTER") == expected_pointer,
+                "For a branch run the driver restart pointer must be set to the "
+                "dated rpointer.cpl of the reference case. Set it with:\n"
+                "  ./xmlchange DRV_RESTART_POINTER={}".format(expected_pointer),
+            )
 
     # determine wav_ice_coupling
-    config["wav_ice_coupling"] = "off"
```
billsacks: Why was this default setting of wav_ice_coupling removed? It seems like it may be needed.

cime_config/buildnml
```diff
@@ -552,6 +600,27 @@ def _create_drv_namelists(case, infile, confdir, nmlgen, files):
...
+            append_line = True
+            with open(user_nl_cpl) as user_file:
+                for line in user_file.readlines():
+                    if 'Flrl_rofsur_nonh2o' in line:
+                        append_line = False
+        if append_line:
+            print("NOTE: Adding the following line to user_nl_cpl")
+            print(   "histaux_lnd2med_file2_flds = \'Flrl_rofsur:Flrl_rofi:Flrl_rofgwl:Flrl_rofsub:Flrl_rofsur_nonh2o\'\n")
+            with open(user_nl_cpl, 'a') as user_file:
+                user_file.write("histaux_lnd2med_file2_flds = \'Flrl_rofsur:Flrl_rofi:Flrl_rofgwl:Flrl_rofsub:Flrl_rofsur_nonh2o\'\n")
```
billsacks: This feels fragile - e.g., duplicating the current list of fields in histaux_lnd2med_file2_flds, which could change over time - and I don't love that buildnml is modifying user_nl_cpl. Can we talk about whether there might be a different way to handle this?

Would it be possible, for example, to add Flrl_rofsur_nonh2o to the standard list of fields, but then have some logic that avoids trying to write a field that isn't connected?

cime_config/config_component.xml
```diff
@@ -39,6 +39,10 @@
     <type>char</type>
     <valid_values>cesm,noresm</valid_values>
     <default_value>cesm</default_value>
+    <values>
+      <value cime_model="cesm">cesm</value>
+      <value cime_model="noresm">cesm</value>
```
billsacks: Is this right? Should COUPLING_MODE be set to noresm if cime_model is noresm?

cime_config/config_component.xml
```diff
@@ -2526,10 +2531,9 @@
     <default_value>FALSE</default_value>
     <group>run_flags</group>
     <file>env_run.xml</file>
-    <desc>turns on coupler bit-for-bit reproducibility with varying pe counts</desc>
```
billsacks: Please revert this change - it will conflict with #704

cime_config/namelist_definition_drv.xml
```diff
@@ -990,14 +1016,13 @@
     <type>char</type>
     <category>budget</category>
     <group>MED_attributes</group>
-    <valid_values>v0,v1</valid_values>
+    <valid_values>v0,v1,v2</valid_values>
     <desc>
-      currently v0 refers to budgets using POP and v1 refers to budgets using MOM6
+      currently v0 refers to budgets using BLOM and v1 refers to budgets using MOM6
```
billsacks: Can you add what v2 refers to?

cime_config/namelist_definition_drv.xml
```diff
@@ -1805,7 +1830,71 @@
     </values>
   </entry>
 
-  
+  <!-- MED histaux lnd2med file 1 - all fields from land with instantaneous output-->
```
billsacks: This comment seems wrong

cime_config/namelist_definition_drv.xml
```diff
@@ -1805,7 +1830,71 @@
...
+    <desc>Auxiliary mediator l2x fields every lnd coupling interval</desc>
+    <values>
+      <value>.false.</value>
+    </values>
+  </entry>
+  <entry id="histaux_lnd2med_file2_flds">
+    <type>char</type>
+    <category>aux_hist</category>
+    <group>MED_attributes</group>
+    <desc>Auxiliary mediator lnd2med output every lnd coupling interval</desc>
```
billsacks: This description (every lnd coupling interval) disagrees with the settings below (daily averages)

mediator/esmFldsExchange_cesm_mod.F90
```diff
@@ -1763,6 +1792,35 @@ subroutine esmFldsExchange_cesm(gcomp, phase, rc)
...
+                  , mrg_type='copy_with_weights', mrg_fracname='ofrac')
+          end if
+          if (fldchk(is_local%wrap%FBImp(compatm, compatm), 'Faxa_hlat', rc=rc) .and. &
+               fldchk(is_local%wrap%FBExp(compocn)         , 'Faxa_hlat', rc=rc)) then
+             call addmap_from(compatm, 'Faxa_hlat', compocn, mapconsf, 'one', atm2ocn_map)
+             call addmrg_to  (compocn, 'Faxa_hlat', mrg_from=compatm ,mrg_fld='Faxa_hlat' &
+                  , mrg_type='copy_with_weights', mrg_fracname='ofrac')
+          end if
+          !if (fldchk(is_local%wrap%FBExp(compatm),'Faxx_hrof', rc=rc)) &
+          !   call addmap_from(compocn, 'Faxx_hrof', compatm, mapconsf, 'one', atm2ocn_map)
```
billsacks: It seems like this commented-out line of code should either be removed or be given some comment as to its future purpose.

mediator/med_diag_mod.F90
```diff
@@ -740,6 +765,7 @@ subroutine diag_atm_send(FB, fldname, nf, areas, lats, afrac, lfrac, ofrac, ifra
     real(r8)               , intent(in)    :: ofrac(:)
     real(r8)               , intent(in)    :: ifrac(:)
     real(r8)               , intent(inout) :: budget(:,:,:)
+    logical                , intent(in)    :: component_contribution
```
billsacks: Can you add a comment documenting the meaning of this component_contribution argument?

mediator/med_phases_history_mod.F90
```diff
@@ -623,6 +627,19 @@ subroutine med_phases_history_write_lnd2glc(gcomp, fldbun_lnd, rc, fldbun_glc)
...
+    if (present(comp_import)) then
+       if (.not. present(fldbun_import)) then
+          call shr_log_error(subname//'if comp_import is present, then fldbun_import must be present', rc=rc)
+          return
+       end if
+       if (comp_import == complnd) then
+          write(hist_file, "(6a)") trim(case_name),'.cpl',trim(inst_tag),'.hx.lnd2glc.',trim(nexttime_str),'.nc'
+       else if (comp_import == compocn) then
+          write(hist_file, "(6a)") trim(case_name),'.cpl',trim(inst_tag),'.hx.ocn2glc.',trim(nexttime_str),'.nc'
+       end if
```
billsacks: This block should have an `else` that aborts with a message like "unhandled comp_import"

mediator/med_phases_history_mod.F90
```diff
@@ -623,6 +627,19 @@ subroutine med_phases_history_write_lnd2glc(gcomp, fldbun_lnd, rc, fldbun_glc)
     if (ChkErr(rc,__LINE__,u_FILE_u)) return
     write(nexttime_str,'(i4.4,a,i2.2,a,i2.2,a,i5.5)') yr,'-',mon,'-',day,'-',sec
     write(hist_file, "(6a)") trim(case_name),'.cpl',trim(inst_tag),'.hx.1yr2glc.',trim(nexttime_str),'.nc'
```
billsacks: Based on my read of the code below, this original setting of hist_file now does nothing, since it will get overwritten. Do you agree? If so, this line should be removed.

mediator/med_phases_history_mod.F90
```diff
@@ -643,18 +660,23 @@ subroutine med_phases_history_write_lnd2glc(gcomp, fldbun_lnd, rc, fldbun_glc)
...
+          ! export field bundle
+          do n = 1,size(fldbun_export)
+             call med_io_write(io_file, fldbun_export(n), whead(m), wdata(m), &
                   is_local%wrap%nx(compglc(n)), is_local%wrap%ny(compglc(n)), &
                   nt=1, pre=trim(compname(compglc(n)))//'Exp', rc=rc)
              if (ChkErr(rc,__LINE__,u_FILE_u)) return
           end do
+       else
+          call shr_log_error(subname//'either fldbun_import or fldbun_export must be present as arguments', rc=rc)
+          return
```
billsacks: This doesn't properly handle the case where both fldbun_import and fldbun_export are present. Should this be handled? If not, there should be an error check of this situation.

mediator/med_phases_history_mod.F90
```diff
@@ -536,20 +539,21 @@ subroutine med_phases_history_write_med(gcomp, rc)
...
 
     ! input/output variables
-    type(ESMF_GridComp)    , intent(in)  :: gcomp
-    type(ESMF_FieldBundle) , intent(in)  :: fldbun_lnd
-    integer                , intent(out) :: rc
-    type(ESMF_FieldBundle) , intent(in), optional :: fldbun_glc(:)
+    type(ESMF_GridComp)               , intent(in)  :: gcomp
+    type(ESMF_FieldBundle) , optional , intent(in)  :: fldbun_import    ! land or ocean import field bundle
+    integer                , optional , intent(in)  :: comp_import      ! either land or ocean component id
+    type(ESMF_FieldBundle) , optional , intent(in)  :: fldbun_export(:) ! export field bundle array
```
billsacks: It looks like it's assumed that fldbun_export is the glc field bundle. This should be clarified in the variable name or at least a comment.

mediator/med_phases_history_mod.F90
```diff
@@ -536,20 +539,21 @@ subroutine med_phases_history_write_med(gcomp, rc)
   end subroutine med_phases_history_write_med
 
   !===============================================================================
-  subroutine med_phases_history_write_lnd2glc(gcomp, fldbun_lnd, rc, fldbun_glc)
+  subroutine med_phases_history_write_data2glc(gcomp, fldbun_import, comp_import, fldbun_export, rc)
```
billsacks: It looks like the changes in this routine will lead to changes in the file names produced. Since I think CESM is regularly producing these aux hist files of fields to glc, it would help to have a summary of how this will change so that people aren't caught by surprise. Can you please add this to the PR description and then I can point some people to this?

mediator/med_phases_prep_rof_mod.F90
```diff
@@ -356,13 +360,14 @@ subroutine med_phases_prep_rof(gcomp, rc)
        if (chkerr(rc,__LINE__,u_FILE_u)) return
     end if
 
-    ! Reset the irrig_flux_field with the map_lnd2rof_irrig calculation below if appropriate
-    if ( NUOPC_IsConnected(is_local%wrap%NStateImp(complnd), fieldname=trim(irrig_flux_field))) then
+    if ( NUOPC_IsConnected(is_local%wrap%NStateImp(complnd), fieldname=trim(irrig_flux_field)) .and. &
+         is_local%wrap%med_coupling_active(comprof,complnd)) then
```
billsacks: I'd like to better understand this change. I'm particularly wondering if there should be an abort if the irrigation field is connected but the rof-lnd coupling is inactive, rather than falling through to the else that sets the irrigation field to 0 in this case

mediator/med_phases_prep_atm_mod.F90
```diff
@@ -233,14 +235,25 @@ subroutine med_phases_prep_atm(gcomp, rc)
       end if
     end do
 
-    ! Add enthalpy correction to sensible heat if appropriate
-    if (FB_FldChk(is_local%wrap%FBExp(compatm), 'Faxx_sen', rc=rc)) then
+    ! Only do the following correction if the mediator is computing the enthalpy to be sent to the ocean
+    ! from rain, snow, etc.
+    if (       FB_FldChk(is_local%wrap%FBExp(compatm), 'Faxx_sen' , rc=rc) .and. &
+         .not. FB_fldchk(is_local%wrap%FBExp(compocn), 'Faxa_hmat', rc=rc)) then
```
billsacks: Can this be made more explicit as to why it's checking for the presence of Faxa_hmat? Is its absence a signal that the mediator is computing the enthalpy to be sent to the ocean? If so, that should at least be stated explicitly in the comment - though the ideal would be to have an explicit flag to this effect, if possible.

mediator/med_phases_prep_atm_mod.F90
```diff
@@ -233,14 +235,25 @@ subroutine med_phases_prep_atm(gcomp, rc)
...
        call FB_getfldptr(is_local%wrap%FBExp(compatm), 'Faxx_sen', dataptr1, rc=rc)
        if (ChkErr(rc,__LINE__,u_FILE_u)) return
        do n = 1,size(dataptr1)
           dataptr1(n) = dataptr1(n) + global_htot_corr(1)
        end do
     end if
+    ! Only do the following if the atmosphere is computing the enthalpy to be sent to the ocean
+    ! from rain, snow, etc.
+    if (FB_FldChk(is_local%wrap%FBExp(compatm), 'Faxx_hrof', rc=rc)) then
```
billsacks: As above: it would be great if the connection between the condition in the `if` was made more explicit.

mediator/med_phases_prep_atm_mod.F90
```diff
@@ -304,4 +319,47 @@ subroutine med_phases_prep_atm_enthalpy_correction (gcomp, hcorr, rc)
...
+    ! Determine sum of enthalpy correction for each hcorr index locally
+    local_hrof_corr(1) = 0._r8
+    do n = 1,size(hcorr)
+       local_hrof_corr(1) = local_hrof_corr(1) + hcorr(n)
+    end do
+    call ESMF_GridCompGet(gcomp, vm=vm, rc=rc)
+    if (ChkErr(rc,__LINE__,u_FILE_u)) return
+    call ESMF_VMAllreduce(vm, senddata=local_hrof_corr, recvdata=global_hrof_corr, count=1, &
+         reduceflag=ESMF_REDUCE_SUM, rc=rc)
+    if (ChkErr(rc,__LINE__,u_FILE_u)) return
```
billsacks: This should be changed to use the new med_global_sums, to ensure bit-for-bit reproducibility with changing processor counts - see https://github.com/ESCOMP/CMEPS/pull/704/changes#diff-52d3090e2ddf0c8ecab551c02b0c20ab8b811548ed902ff2baf8198dbcc2322b

mediator/med_phases_prep_glc_mod.F90
```diff
@@ -637,8 +637,21 @@ subroutine med_phases_prep_glc_avg(gcomp, rc)
...
+    ! -------------------------------------------------
+    write_histaux_l2x1yrg = .false.
+    if (lndAccum2glc_cnt > 0) then
+       call NUOPC_CompAttributeGet(gcomp, name="histaux_l2x1yrg", value=cvalue, &
+            isPresent=isPresent, isSet=isSet, rc=rc)
+       if (ChkErr(rc,__LINE__,u_FILE_u)) return
+       if (isPresent .and. isSet) then
+          read(cvalue,*) write_histaux_l2x1yrg
+       end if
+    end if
```
billsacks: Is there a reason this retrieval of write_histaux_l2x1yrg was moved outside the do_avg conditional block? If I'm reading the code right, it looks like it's only used inside that block. Not a big deal, though....

mediator/med_phases_prep_glc_mod.F90
```diff
@@ -695,62 +711,69 @@ subroutine med_phases_prep_glc_avg(gcomp, rc)
                 where (data2d == 0._r8) data2d = shr_const_spval
              end do
           end do
+
+          ! Write import auxiliary file for ocn if appropriate
+          if (write_histaux_l2x1yrg) then
+             call med_phases_history_write_data2glc(gcomp, fldbun_import=FBocnAccum2glc_o, comp_import=compocn, rc=rc)
```
billsacks: It seems unintuitive that write_histaux_l2x1yrg also controls writing the *ocean* fields. This at the very least deserves a comment calling out this extra use of a variable whose name implies that it's used just for land fields.

mediator/med_phases_prep_ocn_mod.F90
```diff
@@ -112,6 +129,20 @@ subroutine med_phases_prep_ocn_accum(gcomp, rc)
...
+    real(r8), pointer     :: Faxa_hmat (:)
+    real(r8), pointer     :: Faxa_hlat (:)
+    real(r8), allocatable :: hrof2atm(:)
+    real(r8)              :: ocean_htot_corr(1)
+    real(r8)              :: ocean_atot_corr(1)
+    real(r8), allocatable :: hrof(:)
+    ! if separate_varlat is true then do global ocean average for
+    ! hmat_oa only for the net-mass part, and pass in hmat only local
+    ! variable latent heat correction part
+    logical, parameter :: separate_varlat=.true.
```
billsacks: Just checking: is the intent that separate_varlat is a hard-coded thing, not settable via a config setting?

mediator/med_phases_prep_ocn_mod.F90
```diff
@@ -166,100 +197,213 @@ subroutine med_phases_prep_ocn_accum(gcomp, rc)
     if (ChkErr(rc,__LINE__,u_FILE_u)) return
 
     !---------------------------------------
-    !--- custom calculations
+    !--- custom calculation - enthalpy flux computed in mediator for cesm
```
billsacks: These enthalpy flux calculations would ideally be moved to their own subroutine - though that could happen in a later cleanup.

mediator/med_phases_prep_ocn_mod.F90
```diff
@@ -679,4 +823,40 @@ subroutine med_phases_prep_ocn_custom(gcomp, rc)
...
+    ! sum contributions to integral
+    local_sum(1) = 0._r8
+    do n = 1,size(local_array)
+       local_sum(1) = local_sum(1) + local_array(n)
+    end do
+    call ESMF_GridCompGet(gcomp, vm=vm, rc=rc)
+    if (ChkErr(rc,__LINE__,u_FILE_u)) return
+    call ESMF_VMAllreduce(vm, senddata=local_sum, recvdata=global_integral, count=1, &
+         reduceflag=ESMF_REDUCE_SUM, rc=rc)
+    if (ChkErr(rc,__LINE__,u_FILE_u)) return
```
billsacks: This should be replaced by a call to med_global_sums - see https://github.com/ESCOMP/CMEPS/pull/704/changes#diff-52d3090e2ddf0c8ecab551c02b0c20ab8b811548ed902ff2baf8198dbcc2322b

billsacks (changes requested):
Thanks a lot for all of these fixes and enhancements! I have a variety of comments, some trivial and some slightly less trivial. For the less trivial ones, it may be easiest if we talk this through. It will be great to get these CMEPS versions synced up!!

billsacks:
@alperaltuntas - I have gone through this pretty carefully and talked to @mvertens about it. You're welcome to review it if you want, but we feel that you don't have to. So I'm removing you as a requested reviewer, but you can re-add yourself if you want us to wait for your review.

cime_config/buildnml
mvertens: Good point. I have moved this logic to namelist_definition_drv.xml.

cime_config/buildnml
mvertens: Thanks for catching this. I've reintroduced it.

cime_config/buildnml
mvertens: Thanks for raising this as well. Actually it ends up this is not needed.
In med_phases_history_mod already drops requested aux-history fields that aren't connected.
When CLM isn't sending non-water tracers it's never in FBImp(complnd,complnd), so the history filter drops it and the aux file just carries the four water fields. 
So in namelist_definition_drv.xml I now have:
 ```
 <entry id="histaux_lnd2med_file2_flds">
    <type>char</type>
    <category>aux_hist</category>
    <group>MED_attributes</group>
    <desc>Auxiliary mediator lnd2med output every lnd coupling interval</desc>
    <!-- Note that in med_phases_history_mod already drops requested aux-history fields that aren't connected -->
    <!-- When CLM isn't sending non-water tracers it's never in -->
    <!-- FBImp(complnd,complnd), so the history filter drops it and the aux -->
    <!-- file just carries the four water fields. -->
    <values>
      <value>'Flrl_rofsur:Flrl_rofi:Flrl_rofgwl:Flrl_rofsub'</value>
      <value coupling_mode="noresm">'Flrl_rofsur:Flrl_rofi:Flrl_rofgwl:Flrl_rofsub:Flrl_rofsur_nonh2o'</value>
    </values>
  </entry>
```

File: cime_config/config_component.xml (hunk 4 above)
Comment (mvertens) [reply cime_config/config_component.xml]: This was incorrect. Thanks for the catch. Fixed.

File: cime_config/config_component.xml (hunk 5 above)
Comment (mvertens) [reply cime_config/config_component.xml]: Fixed.

File: cime_config/namelist_definition_drv.xml (hunk 6 above)
Comment (mvertens) [reply cime_config/namelist_definition_drv.xml]: I've added comments.

File: cime_config/namelist_definition_drv.xml (hunk 7 above)
Comment (mvertens) [reply cime_config/namelist_definition_drv.xml]: I have removed it.

File: cime_config/namelist_definition_drv.xml (hunk 8 above)
Comment (mvertens) [reply cime_config/namelist_definition_drv.xml]: I've changed the description to 'Auxiliary mediator l2x rof fields averaged over the day.'

File: mediator/esmFldsExchange_cesm_mod.F90 (hunk 9 above)
Comment (mvertens) [reply mediator/esmFldsExchange_cesm_mod.F90]: I've fixed this - but at this point the code to actually compute enthalpy in CAM as never made it to the head of NorESM3 - so in fact this whole block of code is not activated at the moment.

File: mediator/med_diag_mod.F90 (hunk 10 above)
Comment (mvertens) [reply mediator/med_diag_mod.F90]: I have added the following comment:
! Note - if component_contribution is .true. then compute the individual ocn, lnd and ice
! contributions to the budget term

File: mediator/med_phases_history_mod.F90 (hunk 11 above)
Comment (mvertens) [reply mediator/med_phases_history_mod.F90]: I've added an else block and also changed the comment of the subroutine to reflect that there could be ocn input as well:
! Write yearly average of (lnd or ocn)->glc fields on both (lnd or ocn) grids and glc grids

File: mediator/med_phases_history_mod.F90 (hunk 12 above)
Comment (mvertens) [reply mediator/med_phases_history_mod.F90]: Yes - totally agree. Really great catch. Thanks for looking at this so carefully. I've removed that line.

File: mediator/med_phases_history_mod.F90 (hunk 13 above)
Comment (mvertens) [reply mediator/med_phases_history_mod.F90]: Agreed. I've put in an additional error condition if both are present - only one can be present.

File: mediator/med_phases_history_mod.F90 (hunk 14 above)
Comment (mvertens) [reply mediator/med_phases_history_mod.F90]: That is correct. I've clarified this in the comment.

File: mediator/med_phases_history_mod.F90 (hunk 15 above)
Comment (mvertens) [reply mediator/med_phases_history_mod.F90]: I have updated the PR summary.

File: mediator/med_phases_prep_rof_mod.F90 (hunk 16 above)
Comment (mvertens) [reply mediator/med_phases_prep_rof_mod.F90]: I think we need to talk about this.

File: mediator/med_phases_prep_atm_mod.F90 (hunk 18 above)
Comment (mvertens) [reply mediator/med_phases_prep_atm_mod.F90]: I have introduced a new module variable `component_computes_enthalpy_flux` that should help clarify this.

File: mediator/med_phases_prep_atm_mod.F90 (hunk 17 above)
Comment (mvertens) [reply mediator/med_phases_prep_atm_mod.F90]: I have introduced a new module variable `component_computes_enthalpy_flux1 that should help clarify this - and also removed the query from `Faxa_hmat` above.

File: mediator/med_phases_prep_atm_mod.F90 (hunk 19 above)
Comment (mvertens) [reply mediator/med_phases_prep_atm_mod.F90]: Done.

File: mediator/med_phases_prep_glc_mod.F90 (hunk 21 above)
Comment (mvertens) [reply mediator/med_phases_prep_glc_mod.F90]: I agree that this is unintuitive and at the time the ocean input as added as an option the name should have been changed  - but unfortunately for backwards compatibility it was kept. I've added a comment.

File: mediator/med_phases_prep_ocn_mod.F90 (hunk 22 above)
Comment (mvertens) [reply mediator/med_phases_prep_ocn_mod.F90]: Yes - unfortunately that is the case. Since CAM was updated to have the computation of enthalpy for the main NorESM3 development code - the functionality is there in CMEPS to handle the CAM option with some flexibilty - and the goal was to eventually not have this as a parameter at all.

File: mediator/med_phases_prep_ocn_mod.F90 (hunk 24 above)
Comment (mvertens) [reply mediator/med_phases_prep_ocn_mod.F90]: Fixed.

File: cime_config/buildnml (hunk 1 above)
Comment (billsacks) [reply cime_config/buildnml]: Thanks. That's a better place for it.

And just recording what you told me yesterday: It sounds like there's no fundamental reason why histaux_l2x1yrg needs to be true when using blom & clm, but this accommodates a request from NorESM scientists.

File: mediator/med_phases_prep_ocn_mod.F90 (hunk 23 above)
Comment (mvertens) [reply mediator/med_phases_prep_ocn_mod.F90]: I have moved them to their own subroutine and am testing this now.

File: cime_config/config_component.xml [hunk 25]
```diff
@@ -39,6 +39,10 @@
     <type>char</type>
     <valid_values>cesm,noresm</valid_values>
     <default_value>cesm</default_value>
+    <values>
+      <value cime_model="cesm">cesm</value>
+      <value cime_model="noresm">noresm</value>
```
billsacks: Not directly related to this diff block, but somewhat related:

It looks like, in reverting some of your changes in config_component.xml, you also removed this setting for noresm for CASE_SUPPORT_LIBRARIES:

```xml
<value cime_model="noresm">gptl,pio,csm_share,FTorch,CDEPS</value>
```

Was that intentional?

mediator/med_phases_history_mod.F90
```diff
@@ -643,18 +662,28 @@ subroutine med_phases_history_write_lnd2glc(gcomp, fldbun_lnd, rc, fldbun_glc)
...
+          do n = 1,size(fldbun_export)
+             call med_io_write(io_file, fldbun_export(n), whead(m), wdata(m), &
                   is_local%wrap%nx(compglc(n)), is_local%wrap%ny(compglc(n)), &
                   nt=1, pre=trim(compname(compglc(n)))//'Exp', rc=rc)
              if (ChkErr(rc,__LINE__,u_FILE_u)) return
           end do
+       else
+          if (present(fldbun_import) .and. present(fldbun_export)) then
+             call shr_log_error(subname//'fldbun_import and fldbun_export cannot both be present as arguments', rc=rc)
+             return
```
billsacks: I don't think this is the right place for this new check: If both are present, you'll enter the first `if (present(fldbun_import))` block, so I think you need to move this check into that first block - i.e., in the `if (present(fldbun_import))` block, add a check `if (present(fldbun_export))` and raise an error in that case.

mediator/med_phases_prep_atm_mod.F90
```diff
@@ -234,13 +241,42 @@ subroutine med_phases_prep_atm(gcomp, rc)
...
+       else
+          component_computes_enthalpy_flux = 'none'
+       end if
+    end if
+
+    ! Only do the following correction if the mediator is computing the enthalpy to be sent to the ocean
+    ! from rain, snow, etc.
+    ! Note that global_htot_corr is preset to zero as a module variable - and will only be set differently
+    ! if med_phases_prep_atm_enthalpy_correction is called in component_computes_enthalpy_flux == 'med'
+    if (trim(component_computes_enthalpy_flux) /= 'med') then
```
billsacks: Just double-checking this logic: the comment says, "Only do the following if the mediator is computing the enthalpy", but then the logic seems to say the opposite - if component_computes_enthalpy_flux is *not* 'med'. Is this right?

mediator/med_phases_prep_atm_mod.F90
```diff
@@ -234,13 +241,42 @@ subroutine med_phases_prep_atm(gcomp, rc)
...
+          if (ChkErr(rc,__LINE__,u_FILE_u)) return
+          do n = 1,size(dataptr1)
+             dataptr1(n) = dataptr1(n) + global_htot_corr
+          end do
+       end if
+    end if
+
+    ! Only do the following if the atmosphere is computing the enthalpy to be sent to the ocean
+    ! from rain, snow, etc.
+    if (trim(component_computes_enthalpy_flux) /= 'atm') then
```
billsacks: Again, just double-checking this logic: the comment says, "Only do the following if the atmosphere is computing the enthalpy", but then the logic seems to say the opposite - if component_computes_enthalpy_flux is *not* 'atm'. Is this right?

billsacks (changes requested):
Thanks a lot for addressing my comments, @mvertens . I'm happy with nearly everything you've done. I do have a few new comments based on your changes... some of them may be things that are actually correct but I want to double-check with you.

Beyond that, the one remaining thing to discuss is the logic around irrigation in med_phases_prep_rof_mod.

cime_config/config_component.xml
mvertens: No that was definitely not intentional - it's back in now. Thanks for catching that.

mediator/med_phases_history_mod.F90
mvertens: Good point. It's fixed now.

mediator/med_phases_prep_atm_mod.F90
mvertens: The comment is right - my new logic was incorrect. Thanks for catching this.

mediator/med_phases_prep_atm_mod.F90
mvertens: The comment is right - my new logic was incorrect. Thanks for catching this.

billsacks (approved):
This looks good now! As I discussed with @mvertens , I'll open an issue related to the irrigation change - that doesn't need to hold up this PR.

The remaining to do is rerunning the prealpha tests. I'd like to see a full run of the CESM prealphas on both derecho and izumi with all compilers. @mvertens - it sounded like you're planning to at least run the derecho intel tests. Maybe a good plan could be for you to run those, and then we can ask Chris to run prealphas on the other compilers and on izumi just to confirm there are no issues before merging. (I want to be a little more conservative than usual given how close we are to release.)

billsacks:
@DeniseWorthen - would you like to look at this or run any testing on it before it's merged?

DeniseWorthen:
@billsacks Thanks, except for the ocean albedo changes I don't think anything would touch UFS, and from my reading of the code even that should be fine. Let me run a couple of quick tests just to verify.

billsacks:
Thanks a lot for your quick reply @DeniseWorthen !

DeniseWorthen:
@billsacks This appears fine for us; thanks for checking.

DeniseWorthen (approved):

billsacks:
Thanks for confirming @DeniseWorthen !

PR #695, opened 2026-08-17
MAGE coupling

fvitt:
### Description of changes

Split MPI communicator to allow 2-way coupling of WACCM-X with MAGE magnetosphere in a MPMD job.

### Specific notes
See comments added to `cesm/driver/esmApp.F90`

Contributors other than yourself, if any: @phamkh

bit-for-bit

N/A

### Testing performed

aux_cam on izumi and derecho

### DIFF
--- a/cesm/driver/esmApp.F90
+++ b/cesm/driver/esmApp.F90
@@ -41,6 +41,8 @@ program esmApp
 #else
   call MPI_init(rc)
 #endif
+  ! note: this is superseded by the MPI_COMM_split below, which restricts this
+  ! application to its own subset of MPI_COMM_WORLD
   COMP_COMM = MPI_COMM_WORLD
 
   !-----------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


cesm/driver/esmApp.F90
```diff
@@ -41,6 +41,8 @@ program esmApp
 #else
   call MPI_init(rc)
 #endif
+  ! note: this is superseded by the MPI_COMM_split below, which restricts this
+  ! application to its own subset of MPI_COMM_WORLD
   COMP_COMM = MPI_COMM_WORLD
```
billsacks, line 46: Is there any value in maintaining this line given that it is superseded below?

cesm/driver/esmApp.F90
```diff
@@ -49,6 +51,27 @@ program esmApp
...
+  ! because the cross-model coupling in CAM (src/ionosphere/waccmx/mage_module.F90)
+  ! uses it directly as the pool from which the WACCM-X <-> REMIX coupling
+  ! communicator is built.
+  !
+  ! The color (67) must match myAppId in mage_module.F90; the other models split
+  ! the same world with their own IDs.  A key of 0 preserves the existing rank
+  ! ordering.  In a standalone CESM run every rank passes the same color, so the
+  ! result is simply a duplicate of MPI_COMM_WORLD and behavior is unchanged.
+  !-----------------------------------------------------------------------------
+  call MPI_COMM_split(MPI_COMM_WORLD,67,0,COMP_COMM,ier)
```
billsacks, line 73: According to the documentation of this MPI routine, it looks like this does an allgather. Is that a performance concern? If we keep this method, I'm wondering if we should put it in a conditional on some flag that is only enabled when it's actually needed to avoid this unnecessary complexity and allgather. Something like:

```Fortran
if (running_mpmd) then
   call MPI_COMM_split(MPI_COMM_WORLD,67,0,COMP_COMM,ier)
else
   COMP_COMM = MPI_COMM_WORLD
end if
```

(Note that that block would also replace the above setting of COMP_COMM = MPI_COMM_WORLD a few lines above... it assumes that we don't need an initial setting of COMP_COMM = MPI_COMM_WORLD if we're about to call MPI_COMM_split; I don't know if that's the case or not.)

(The name of the logical – running_mpmd – may not be the best / most appropriate.)

I'm also wondering if, for this relatively simple situation, there might be a better routine to use than MPI_COMM_split, which seems overly general for this scenario. Can this be done via MPI_Comm_create? I ask this totally naively, since I don't really understand this.

billsacks (commented):
Based on input from @cacraigucar , this is connected with https://github.com/escomp/cam/pull/1649. The initial review of that CAM PR has questioned whether we actually want to take the current approach. I'm going to mark this as a draft until this decision is resolved.

In the meantime, I had already reviewed this code and had some questions / thoughts. I'm sharing these, but these don't need any action unless / until we decide to move ahead with this.

PR #696, opened 2026-08-22
CO2A for cplhist in DATM with CLM

ekluzek:
### Description of changes

Set CCSM_BGC to CO2A for DATM CPLHIST forcing mode with CLM

### Specific notes

   Yes, for DATM%CPLHIST mode compsets 

   Rather than constant CO2 CPLHIST cases will use CO2 from the input CPLHIST case driving it

### Testing performed

Running: aux_clm, aux_cdeps testing in ctsm5.4.054 (https://github.com/ESCOMP/CTSM/pull/4180)


cime_config/config_component_cesm.xml
```diff
@@ -108,6 +108,7 @@
       <value compset="_CAM">CO2A</value>
       <value compset="_DATM">none</value>
       <value compset="_DATM%CPLHIST.+BLOM%ECO">CO2A</value>
+      <value compset="_DATM%CPLHIST_CLM60">CO2A</value>
```
ekluzek: OK, this needs to be consistent in DATM, CMEPS and CLM and should be anytime DATM has CPLHIST on, CO2 is used from the input case.

ekluzek: Oh, wait -- how do other components want it set?

ekluzek: For now let's change this to any flavor of CLM.

Later we'll do changes to make things consistent across components. As a set of coordinated PR's across components.

cime_config/config_component_cesm.xml
```diff
@@ -108,6 +108,7 @@
       <value compset="_CAM">CO2A</value>
       <value compset="_DATM">none</value>
       <value compset="_DATM%CPLHIST.+BLOM%ECO">CO2A</value>
+      <value compset="_DATM%CPLHIST_CLM">CO2A</value>
```
billsacks: For slightly more robustness, and for consistency with other matches, this should be:


```suggestion
      <value compset="_DATM%CPLHIST.*_CLM">CO2A</value>
```

(The addition of `.*` handles: (1) If the order were ever allowed to change, so CLM isn't the next component after DATM; and (2) If there could ever be a second modifier after CPLHIST.)

(I am once again reminded of https://github.com/ESCOMP/CMEPS/issues/640.)

billsacks (changes requested):
Thanks a lot for fixing this, @ekluzek !

I have one request for an improvement to the regex match.

Beyond that, I just want to confirm: With this addition, we'll use CO2A for *any* CPLHIST run with CLM, regardless of whether it's transient or spinup, but only for *transient* non-CPLHIST datm-forced runs (HIST or SSP). Is that the intended behavior?

cime_config/config_component_cesm.xml
ekluzek: Thanks for that point. I should do something similar in other places in my CTSM PR where I'm doing things like this.

ekluzek: Did this in e4aa62d9

ekluzek:
> Thanks a lot for fixing this, @ekluzek !
> 
> I have one request for an improvement to the regex match.

Yes, did that.

> 
> Beyond that, I just want to confirm: With this addition, we'll use CO2A for _any_ CPLHIST run with CLM, regardless of whether it's transient or spinup, but only for _transient_ non-CPLHIST datm-forced runs (HIST or SSP). Is that the intended behavior?

Yes, that's correct. It's just adding the bit about CPLHIST with CLM for any mode. This is the restricted change I want to make now. The idea is that the CO2 on the CPLHIST files is the truth about what CO2 should be for any CPLHIST case trying to duplicate it. The CO2 from CAM can be different from DATM (CAM has latitude bands), and CAM might've had prognostic CO2, or non-standard settings or datasets. So using CO2 from it as truth is the best default to use.

We are also going to need to do something for MOM with MARBL, but we need to talk to @mnlevy1981 and @alperaltuntas about that. And that will be a future PR.

The more general solution is to do #698 which I think is clearer and more robust, but requires changes across: CMEPS, CDEPS, CTSM, MOM, and CAM so we need to decide if/when that should be done.

ekluzek:
OK, I'm running aux_cdeps and aux_clm testing with this now, and if that looks good I'd like to merge.

cime_config/config_component_cesm.xml
```diff
@@ -108,6 +108,7 @@
       <value compset="_CAM">CO2A</value>
       <value compset="_DATM">none</value>
       <value compset="_DATM%CPLHIST.+BLOM%ECO">CO2A</value>
+      <value compset="DATM%CPLHIST.*_CLM">CO2A</value>
```
billsacks: Please change this to restore the leading `_` that you removed:

```suggestion
      <value compset="_DATM%CPLHIST.*_CLM">CO2A</value>
```

In practice it probably doesn't matter, but I find these arbitrary inconsistencies in regexes to be a source of confusion and possible bugs.

billsacks (changes requested):
Thanks for the change, but in making that change you introduced another change that I'd like you to revert:

billsacks (approved):
Thanks, looks good now!

@ekluzek - I'm happy to merge this if you feel it's ready - please let me know if I should go ahead and merge.

ekluzek:
@billsacks could you go ahead and do the merge now? I haven't had a reason to change this for quite a while now. This part is already tested and it's dead simple.

Issue #483, opened 2024-07-10
Runtime hang when ESMF_AWARE_THREADING is on


nmizukami:
Related to #482. 
Once two compiling error described in #482 locally, the model is successfully compiled. when try running, however, it hangs at very beginning of the initialization. I got PET error like this.

```
20240708 185822.851 ERROR            PET1290 esm.F90:1157 Not valid  -  Component ROF has mpi task mismatch, do threads align with nodes?
20240708 185822.855 ERROR            PET1290 esm.F90:203 Not valid  - Passing error in return code
20240708 185822.855 ERROR            PET1290 ESM0001:src/addon/NUOPC/src/NUOPC_Driver.F90:797 Not valid  - Passing error in return code
20240708 185822.855 ERROR            PET1290 ensemble:src/addon/NUOPC/src/NUOPC_Driver.F90:2901 Not valid  - Phase 'IPDv02p1' Initialize for modelComp 1: ESM0001 did not r
eturn ESMF_SUCCESS
20240708 185822.855 ERROR            PET1290 ensemble:src/addon/NUOPC/src/NUOPC_Driver.F90:1329 Not valid  - Passing error in return code
20240708 185822.855 ERROR            PET1290 ensemble:src/addon/NUOPC/src/NUOPC_Driver.F90:486 Not valid  - Passing error in return code
20240708 185822.855 ERROR            PET1290 esmApp.F90:134 Not valid  - Passing error in return code
20240708 185822.855 INFO             PET1290 Finalizing ESMF
```
This needs some esmf expert knowledge to look deeper into the problem.

PR #700, opened 2026-09-07
Small fixes to address some issues with ESMF managed threading

briandobbins:
### Description of changes

In CESM, ESMF_AWARE_THREADING (ESMF's managed threading) results in hangs because of issues with how PEs that are threads are treated like full PETs in CMEPS & NUOPC at the moment, but they can't do things like query the component VM.  Also, connectors built on PE lists (vs PET lists) would hang during RouteHandle creation.

The primary addition here is a small helper function that identifies PET-owning PEs (via a strided ID through an array of all PEs).  This might be better done inside ESMF, and if so, this code can be simplified in the future, but should be able to co-exist with those changes too.

### Specific notes

This is not a codebase I'm very familiar with, and a lot of this was aided by Claude.  I have run some test with it, but not an exhaustive CESM testlist yet.  I'd love feedback and to better understand the process for testing this in full, if not just via a new CESM tag pulling this in.

Claude AI

#701 
Potentially #483 but this was done independently and related to needing threaded runs for CMIP7 with CESM3.  No additional issue was created for that yet, as this is still a work in progress.

No

No

### Testing performed
Simple tests with a FHISTC_MTt4s case only for now, on Derecho -- expecting to do a coupled case later today.

### DIFF
--- a/cesm/nuopc_cap_share/driver_pio_mod.F90
+++ b/cesm/nuopc_cap_share/driver_pio_mod.F90
@@ -187,7 +187,7 @@ contains
     integer, intent(out) :: rc
 
     type(ESMF_VM) :: vm
-    integer :: i, npets, default_stride
+    integer :: i, npets, default_stride, lpet
     integer :: j, myid
     integer :: k
     integer :: comp_comm, comp_rank
@@ -275,6 +275,17 @@ contains
        if(associated(gcomp)) then
           petlocal(i) = ESMF_GridCompIsPetLocal(gcomp(i), rc=rc)
           if (chkerr(rc,__LINE__,u_FILE_u)) return
+
+          ! Under ESMF-managed threading, PETs whose PEs are lent to another PET's
+          ! threads hold no VM for the child; ESMF_VMGet returns localPet = -1 there
+          ! (NUOPC refdoc sec. 2.7.2), so treat them as non-local for PIO setup.
+          if (petlocal(i)) then
+             call ESMF_GridCompGet(gcomp(i), vm=vm, rc=rc)
+             if (chkerr(rc,__LINE__,u_FILE_u)) return
+             call ESMF_VMGet(vm, localPet=lpet, rc=rc)
+             if (chkerr(rc,__LINE__,u_FILE_u)) return
+             petlocal(i) = (lpet >= 0)
+          end if
 
           call NUOPC_CompAttributeGet(gcomp(i), name="pio_async_interface", value=cval, rc=rc)
           if (chkerr(rc,__LINE__,u_FILE_u)) return


theurich:
@briandobbins I took a quick look at this, trying to figure out what the exact symptom in CESM is that triggered this PR. From an ESMF perspective, the way petLists are handled currently is the "correct" way. With ESMF-managed threading, components using the same resources are allowed to use different threading levels, in fact that is one of the features of ESMF-managed threading. There should be no hangs due to that if used correctly.

My sense is that this PR might work-around some deeper issue, either a bug in ESMF, or some sort of incorrect use of the feature in CESM. I don't doubt that the work-around helps move things forward, but I don't think it is what we ultimately want. E.g. NUOPC_Connectors should work just fine on the union of petLists given to components that use them to do some threading on them, while the NUOPC_Connector runs unthreaded on the union. This is tested and has been exercised in UFS, etc.

With that said, can you point me to the specific issues with hangs in CESM/CMEPS? Maybe that will shed some more light on what is going on here. Thanks.

briandobbins:
@theurich Thanks for your insight!  This is super helpful -- if the behavior is working as intended, then this must be an issue with how CMEPS sets up our layouts (in esm_init_pelayout in esm.F90), and that makes for a much simpler PR.  I think I can use vmIsPresent to gate the ESMF_GridCompGet call, since that was part of the problem -- PEs that were being used as threads didn't have a valid VM (nullpointer).

My guess is there's still a _real_ bug in the ESMF_VMIsCreated returning a null pointer on PEs that are just threads, but this can likely be fixed purely in CMEPS now.  I'll give this a shot when I'm out of meetings and report back.

briandobbins:
@billsacks @mvertens  OK, I've updated this in a simple way -- now we query the ESMF_VMGetThis function to determine if a VM has a null pointer (as happens to PE that is merged into another PET's set of PEs for threading).   Tested on a threaded case, with more testing to come, but I feel a _lot_ happier about this.

briandobbins:
@theurich Technically, I think there's still a bug in ESMF, albeit a much smaller one with a one-line fix -- let me know what you think.  Basically, ESMF_VMisCreated returns true _even if_ the VM pointer is null.  Right now, the check is just this:

```
 if (ESMF_VMGetInit(vm)==ESMF_INIT_CREATED) &
      ESMF_VMIsCreated = .true.
```

Permalink:
https://github.com/esmf-org/esmf/blob/4161bf9d4493b8e909e0f44d93581f326e304a7d/src/Infrastructure/VM/interface/ESMF_VM.F90#L5830

But I would think this is more 'correct', since it isn't a VM if it has a null pointer:

```
if (ESMF_VMGetInit(vm)==ESMF_INIT_CREATED .and. vm%this /= ESMF_NULL_POINTER) &
    ESMF_VMIsCreated = .true.

```

Thoughts?

mvertens:
@briandobbins - the new changes are a big improvement. However, since they are now always invoked - I would run at least a prealpha suite to make sure that this does not have any unintended behavior for non-threaded implementations.

briandobbins:
@mvertens Sounds good, will do.  Thank you!

cesm/driver/esm.F90
```diff
@@ -1161,12 +1163,18 @@ subroutine esm_init_pelayout(driver, maxthreads, rc)
 
        if (ESMF_GridCompIsPetLocal(child, rc=rc)) then
 
+          ! Under ESMF-managed threading, PEs that only contribute cores to another PET
+          ! hold no VM for the component; its internal pointer is null there, so skip them.
```
billsacks: Thanks for adding this explanatory comment. Can you move it a couple of lines down, so it is more clearly associated with the new ESMF_VMGetThis call and the following conditional on vmThis?

cesm/driver/esm.F90
billsacks, line 1182: To try to understand the possible impact of the above changes, I was trying to figure out the purpose of this conditional (which seems to be the reason why we need to get comms(i+1) in the first place). My impression is that this is just here for the error check (`nprocs /= ntasks`) and other than that, this doesn't do anything? Do you agree? If so, the above change seems lower risk.

Furthermore, it looks to me like there is some dead code here (pre-existing - not a new issue in this PR), and I'm curious if it looks the same to you, too. I'm fine with deferring its removal (especially given how close we are to release, so wanting to be lower risk), but just to note it:
- comp_comm_iam appears to be filled but never used
- comp_iamin appears to be filled but never used
- rank appears to be fetched but never used

cesm/driver/esm.F90
```diff
@@ -1161,12 +1163,18 @@ subroutine esm_init_pelayout(driver, maxthreads, rc)
...
           call ESMF_GridCompGet(child, vm=vm, rc=rc)
           if (chkerr(rc,__LINE__,u_FILE_u)) return
-
-          call ESMF_VMGet(vm, mpiCommunicator=comms(i+1), rc=rc)
+          call ESMF_VMGetThis(vm, vmThis, rc=rc)
```
billsacks: It looks like this subroutine isn't part of the ESMF public API, so it's risky to rely on it. My impression is that you need this as a workaround, but I would add documentation calling this out, ideally with a reference to an ESMF issue, so that we know when we can remove this workaround (when the ESMF issue is resolved).

cesm/nuopc_cap_share/driver_pio_mod.F90
```diff
@@ -276,6 +278,16 @@ subroutine driver_pio_component_init(driver, inst_comm, asyncio_petlist, rc)
           petlocal(i) = ESMF_GridCompIsPetLocal(gcomp(i), rc=rc)
           if (chkerr(rc,__LINE__,u_FILE_u)) return
 
+          ! Under ESMF-managed threading, PEs that only contribute cores to another PET
+          ! hold no VM for the component; treat them as non-local for PIO setup.
+          if (petlocal(i)) then
+             call ESMF_GridCompGet(gcomp(i), vm=vm, rc=rc)
+             if (chkerr(rc,__LINE__,u_FILE_u)) return
+             call ESMF_VMGetThis(vm, vmThis, rc=rc)
```
billsacks: As above: I would call out that we're using an ESMF routine that isn't part of the public API here, ideally with a reference to an ESMF issue that talks about resolving the issue on the ESMF side.

billsacks (changes requested):
Thanks a lot for iterating on this, @briandobbins - I'm much happier with this new version. I have some requests related to comments, plus a question that I'd like your confirmation on to help me understand the potential impacts of this change. After that, I'm happy with this coming in after it has undergone testing as suggested by @mvertens .

billsacks:
@briandobbins - Given that it sounds like (at least part of) this is a workaround for an ESMF issue, and it relies on an ESMF routine that appears to be outside the public API, I feel like it would be important to open a CMEPS issue that describes:

(1) What fix in ESMF would allow this change to be backed out

(2) Which portions of this PR could be backed out

(3) How can we test to verify that things are working after backing out these changes and pointing to an ESMF that is supposed to fix the issue

theurich:
Quick note that it is the correct ESMF behavior for the child VM to return `IsCreated==.true.` on all PETs. It just means that the object can be used in calls that take a VM object.

The ESMF prescribed way to determine whether the current (driver) PET is active/inactive under a child component is to query the child VM for `localPet`. If it returns `>=0`, then it is active, and if `==-1`, then it is inactive under the child component.

There is a tiny bit information about this under the `ESMF_VMGet()` API doc, for the `localPet` argument. However, I realize that this needs more explanation. So I just added a new sub-section to the NUOPC doc under the "Resource Control and Threaded Components" section:

https://earthsystemmodeling.org/docs/nightly/feature/nuopc-thread-doc/NUOPC_refdoc/node3.html#SECTION00037200000000000000

Hopefully this provides sufficient clarification. Let me know. Thanks.

billsacks:
@theurich - thanks a lot for this input!

@briandobbins - do you feel it will work to revise this based on @theurich 's input? If I understand the documentation right, it sounds like this might make the change in esm.F90 unnecessary, since the mpiCommunicator should return MPI_COMM_NULL, which is checked for already - though I'm not sure if I'm understanding all of this right. And then it sounds like the pattern in driver_pio_mod can be changed to use the officially supported mechanism for this??

briandobbins:
@billsacks Probably.  Let me take a look tonight -- I didn't love the idea of using a non-public API call either, and since he further clarified that the ESMF_VMIsCreated returning true even when the pointer is null is working as intended, it probably does need either another approach, _or_, we proceed with this approach for now and create an issue for later to find a fix.

If I can test and validate an approach that doesn't use it tonight, great.  If not, then given the pressures to get this in a tag by next week, I'd suggest we merge this and make an issue to find a fix (which would be bit for bit, obviously) prior to a release tag for CESM.

briandobbins:
OK, this has been a 'fun' adventure in trying to do complex things fast while working with an AI -- it helped with a solution that _worked_, but was far from optimal.  It turns out the changes to esm.F90 weren't even needed - this is entirely in how PIO is set up, and that has a relatively simple fix that uses public ESMF API calls.

This new minimal fix works in both the non-threaded and threaded test case I ran.  I'll definitely run it through more testing soon, too.

I probably should've known better.  Reminds me of the "Faster, Cheaper, Better -- pick _two_!" stuff that happened with NASA.  Thank you @theurich @billsacks @mvertens for being extremely helpful voices of reason, and pushiing back on this - sorry to have taken your time when I should've done a wee bit more due diligence first!  Lessons learned for sure.

billsacks (approved):
This looks even better now - thanks!

Given how limited the changes are now, I'd be comfortable with just something like aux_cime_baselines being run on this.

billsacks:
@briandobbins has run aux_cime_baselines on this, and it looks good.

PR #702, opened 2026-09-08
Fix aux history field filter skipping entries

alperaltuntas:
The filter that removes unrealized fields from an aux history field list deletes an entry by shifting the rest of the list down one slot, but it still advances the loop index, so the entry moved into the vacated slot never gets checked. This algorithm doesn't work when two or more consecutive fields are unavailable. The aux history write then asks FBImp for a field that isn't there, and ESMF aborts with "key does not exist".

This PR replaces the loop with a single pass compaction with separate read and write indices, so every entry is checked exactly once. 

testing: aux_ww3


billsacks (approved):
This is a cleaner implementation, in addition to fixing the issue - thanks!

This new algorithm *might* deserve an explanatory comment, since (at least for me) it takes a little thought to convince myself that it's correct and safe - in particular, since the same fieldnamelist is both read-from and written-to. I don't feel strongly about this, so am curious what you think. Things that I think could be worth calling out:
- There is a loop invariant: at the start of each loop iteration, elements 1..fieldCount are fields that we want to include
- At the point of the assignment `fieldnamelist(fieldCount) = fieldnamelist(n)` it is guaranteed that fieldCount <= n, so we do not risk overwriting a still-to-be-checked entry

alperaltuntas:
Just added an explanatory comment. Thanks!

PR #703, opened 2026-09-10
Separate namelist options for removing negative liq vs ice runoff

billsacks:
### Description of changes

At one point this was desired for the CESM3 runs, so that we could remove negative liquid fluxes but maintain negative ice fluxes, so that we could have closer to net-zero ice fluxes into the arctic ocean rather than having ice fluxes biased positive. (These ice fluxes into the arctic ocean are driven largely from CTSM's dynamic landunit conservation corrections, so they should average to around 0, but by removing negative fluxes, we were ending up with positive net fluxes regionally.)

It seems like this is no longer desired, but I'm putting the option in place in case people want to continue to experiment with this.

See https://github.com/NCAR/cesm_dev/discussions/393 for more details.

### Specific notes

Are changes expected to change answers? (specify if bfb, different at roundoff, more substantial) bfb

Any User Interface Changes (namelist or namelist defaults changes)? Splits remove_negative_runoff_lnd into remove_negative_runoff_lnd_liq and remove_negative_runoff_lnd_ice

### Testing performed

CESM's aux_cime_baselines on derecho, in the context of cesm3_0_beta09.

Also ran one-off tests of `SMS_D_Ld2.f19_g17.X.green_gnu.drv-glc_avg_frequently` with the four possible permutations of the two new namelist flags, comparing against baselines generated by changing the code and/or the old namelist flag (as needed), and ensuring bit-for-bit.


mvertens (approved):
LGTM

PR #704, opened 2026-09-17
Use reproducible global sums when BFBFLAG is set

### LINKED ISSUE #650: Leverage BFBFLAG to use reproducible sums in place of ESMF_VMAllreduce
As pointed out by @Katetc - we have a number of places in CMEPS where we use ESMF_VMAllreduce to do global sums. She found this in the course of testing answer changes with changing processor counts for T compsets (CISM forced by DLND), and for that use case, the issue is the global sums in the SMB renormalization (she says that turning off SMB renormalization allows answers to be bit-for-bit when changing processor counts - i.e., glc_renormalize_smb). But, from a quick look through the CMEPS code, it looks like this also comes up in post_rof_mod (dealing with negative runoff) and in prep_atm_mod (enthalpy correction).

I'm not sure we want to go so far as to require that global sums always be bit-for-bit with changing processor count, since that could impact performance. (My impression is that this isn't actually a CESM-wide requirement, but some components have it as a requirement. I'm not sure I'm right about that, though.) But it seems like we should at least have an option for reproducible global sums if a flag is set, and we can then use that flag in our testing. We have a currently-deprecated BFBFLAG, and the easiest course may be to bring that flag back into operation and use it for this purpose.

So I propose:
- Use BFBFLAG to set a nuopc runtime variable that controls whether we do more expensive global reductions that are bit-for-bit across processor counts
- If this flag is set, we can use the reproducible sums in https://github.com/ESCOMP/CESM_share/blob/main/src/shr_reprosum_mod.F90 in place of ESMF_VMAllreduce

@briandobbins @mvertens - also looping you in here; I'd welcome your thoughts.

billsacks:
### Description of changes

There are a few places in CMEPS where global sums are performed. These global sums can be sensitive to processor count. This causes failures in tests - like the ERP test - that check for bit-for-bit results with changing processor count.

This PR resurrects the BFBFLAG xml variable and uses it to choose whether to perform global sums in the original way or using shr_reprosum_mod to get bit-for-bit results with changing processor count. (This is done via a bfbflag attribute; this attribute is in Allcomp so CDEPS can also leverage it.) By default we still use the original way, both for better performance and because UFS doesn't have shr_reprosum_mod. (Note that an error will be raised if trying to set bfbflag with UFS.) Some CIME tests, such as the ERP test, turn on BFBFLAG so will leverage this new behavior.

This is implemented in three places: prep_glc (for SMB renormalization), post_rof (for removal of negative runoff) and prep_atm (for an enthalpy correction). There is one other possibly relevant global sum in CMEPS - for med_diag_mod - which I am *not* handling here; I think this is just for budget diagnostics, which I feel are okay to differ by roundoff with processor changes, and from a quick look, I think that fixing that would require more extensive changes.

Note that, in order to leverage the new routine that computes global sums in either the old or new way (depending on bfbflag), I needed to change the code to put the summands in arrays rather than computing local sums, because the reproducible sum routine requires us to not do any pre-summation on local processors. (An exception is summation across elevation classes on a single grid cell: that summation is safe to do ahead of time since it doesn't depend on processor count.)

Most of the changes were written by Claude, but with detailed guidance and careful reviews by myself.

### Specific notes

- Resolves ESCOMP/CMEPS#650 

Are changes expected to change answers? (specify if bfb, different at roundoff, more substantial) - roundoff-level differences for some configurations:
- Potential roundoff-level changes from baseline for all configurations with DGLC or CISM (due to splitting the sum over elevation classes and the sum over grid cells into two different steps)
- Potential roundoff-level changes from baseline for tests that turn on BFBFLAG (ERP, PEM, PET, SEQ), since this flag is now enabled whereas previously it did nothing

Any User Interface Changes (namelist or namelist defaults changes)? Adds bfbflag attribute

### Testing performed
aux_cime_baselines and aux_cdeps in the context of cesm3_0_beta09, with this branch and the associated CDEPS branch.

All tests pass and are bit-for-bit except expected failures and this test that has roundoff-level changes: `SMS_Ly2.f09_g17_gris20.T1850Gg.derecho_intel`. By putting in place some temporary changes, I verified that these roundoff-level changes are due to the splitting of sums over elevation classes and sums over grid cells in the new code, changing the order of operations from before.

Also some additional tests to give better test coverage; these included the following (along with some others that were less good tests, so I'm not documenting them here)
- SMS_Ld3.ne30pg3_t232.B1850C_LTso.derecho_intel.drv-glc_avg_frequently - bit-for-bit with baseline
- ERP_P64x2_Ld366.f10_f10_mg37.I2000Clm50BgcCrop.derecho_intel.clm-irrig_alternate_monthly - roundoff-level diffs from baseline; bit-for-bit with baseline if I set BFBFLAG FALSE.

I also ran some tests with BFBFLAG TRUE vs. FALSE to confirm that we get only roundoff-level diffs from setting BFBFLAG:
```
SMS_D_Ld3.f19_g17.X.green_gnu.drv-glc_avg_frequently
SMS_Ld3.f19_g17.X.green_gnu.drv-glc_avg_frequently
SMS_D_Ld3.f19_g17.X.green_gnu.drv-water_tracers--drv-glc_avg_frequently
SMS_Ld3.f19_g17.X.green_gnu.drv-water_tracers--drv-glc_avg_frequently
```

For all of these, I made changes (from a branch that will be coming in soon) that change xrof to send negative as well as positive runoff (to better exercise the post_rof code). For the water_tracers tests, I also had to put this in user_nl_cpl:
```
water_tracers_variables_not_checked = "FBExprof%Flrl_irrig_wtracers:FBExpocn%Foxx_rofi_wtracers:FBExpocn%Foxx_rofl_wtracers"
```

These tests showed just roundoff-level diffs between BFBFLAG true and false, as expected.

### DIFF
--- a/cime_config/config_component.xml
+++ b/cime_config/config_component.xml
@@ -1107,6 +1107,15 @@
     <desc>pes or cores used relative to MAX_MPITASKS_PER_NODE for accounting (0 means TOTALPES is valid)</desc>
   </entry>
 
+  <entry id="BFBFLAG">
+    <type>logical</type>
+    <valid_values>TRUE,FALSE</valid_values>
+    <default_value>FALSE</default_value>
+    <group>run_flags</group>
+    <file>env_run.xml</file>
+    <desc>If TRUE, perform CMEPS calculations such as global sums in a manner that is bit-for-bit independent of processor count</desc>
+  </entry>
+
   <!-- ===================================================================== -->
   <!-- definitions grid -->
   <!-- ===================================================================== -->
@@ -2514,21 +2523,6 @@
    <file>env_run.xml</file>
    <desc>add aoflux calculation to runseq</desc>
  </entry>
-
-  <!-- ===================================================================== -->
-  <!-- TODO: the following should be deleted. They are still used by  -->
-  <!-- the component testing but are not doing anything in cmeps -->
-  <!-- ===================================================================== -->
-
-  <entry id="BFBFLAG">
-    <type>logical</type>
-    <valid_values>TRUE,FALSE</valid_values>
-    <default_value>FALSE</default_value>
-    <group>run_flags</group>
-    <file>env_run.xml</file>
-    <desc>turns on coupler bit-for-bit reproducibility with varying pe counts</desc>
-  </entry>
-
 
  <help>
     =========================================
--- a/cime_config/namelist_definition_drv.xml
+++ b/cime_config/namelist_definition_drv.xml
@@ -878,6 +878,19 @@
     </values>
   </entry>
 
+  <entry id="bfbflag" modify_via_xml="BFBFLAG">
+    <type>logical</type>
+    <category>control</category>
+    <group>MED_attributes</group>
+    <desc>
+      If true, perform CMEPS calculations such as global sums in a manner that is
+      bit-for-bit independent of processor count
+    </desc>
+    <values>
+      <value>$BFBFLAG</value>
+    </values>
+  </entry>
+
   <entry id="dbug_flag">
     <type>integer</type>
     <category>expdef</category>
--- a/mediator/med_phases_prep_glc_mod.F90
+++ b/mediator/med_phases_prep_glc_mod.F90
@@ -1277,12 +1277,12 @@ contains
     deallocate(local_accum_lnd)
     deallocate(local_ablat_lnd)
     if (maintask) then
-       write(logunit,'(a,d21.10)') trim(subname)//'global_accum_lnd = ', global_accum_lnd(1)
-       write(logunit,'(a,d21.10)') trim(subname)//'global_ablat_lnd = ', global_ablat_lnd(1)
+       write(logunit,'(a,d28.17)') trim(subname)//'global_accum_lnd = ', global_accum_lnd(1)
+       write(logunit,'(a,d28.17)') trim(subname)//'global_ablat_lnd = ', global_ablat_lnd(1)
        do t = 1, num_wtracers
-          write(logunit,'(a,i0,a,d21.10)') trim(subname)//'tracer #', t, &
+          write(logunit,'(a,i0,a,d28.17)') trim(subname)//'tracer #', t, &
                ': global_accum_lnd = ', global_accum_lnd(1+t)
-          write(logunit,'(a,i0,a,d21.10)') trim(subname)//'tracer #', t, &
+          write(logunit,'(a,i0,a,d28.17)') trim(subname)//'tracer #', t, &
                ': global_ablat_lnd = ', global_ablat_lnd(1+t)
        end do
     endif
@@ -1321,12 +1321,12 @@ contains
     deallocate(local_accum_glc)
     deallocate(local_ablat_glc)
     if (maintask) then
-       write(logunit,'(a,d21.10)') trim(subname)//'global_accum_glc = ', global_accum_glc(1)
-       write(logunit,'(a,d21.10)') trim(subname)//'global_ablat_glc = ', global_ablat_glc(1)
+       write(logunit,'(a,d28.17)') trim(subname)//'global_accum_glc = ', global_accum_glc(1)
+       write(logunit,'(a,d28.17)') trim(subname)//'global_ablat_glc = ', global_ablat_glc(1)
        do t = 1, num_wtracers
-          write(logunit,'(a,i0,a,d21.10)') trim(subname)//'tracer #', t, &
+          write(logunit,'(a,i0,a,d28.17)') trim(subname)//'tracer #', t, &
                ': global_accum_glc = ', global_accum_glc(1+t)
-          write(logunit,'(a,i0,a,d21.10)') trim(subname)//'tracer #', t, &
+          write(logunit,'(a,i0,a,d28.17)') trim(subname)//'tracer #', t, &
                ': global_ablat_glc = ', global_ablat_glc(1+t)
        end do
     endif
@@ -1500,13 +1500,13 @@ contains
     endif
     if (maintask) then
        if (present(tracer_num)) then
-          write(logunit,'(a,i0,a,d21.10)') trim(caller_subname)//'tracer #', tracer_num, &
+          write(logunit,'(a,i0,a,d28.17)') trim(caller_subname)//'tracer #', tracer_num, &
                ': accum_renorm_factor = ', accum_renorm_factor
-          write(logunit,'(a,i0,a,d21.10)') trim(caller_subname)//'tracer #', tracer_num, &
+          write(logunit,'(a,i0,a,d28.17)') trim(caller_subname)//'tracer #', tracer_num, &
                ': ablat_renorm_factor = ', ablat_renorm_factor
        else
-          write(logunit,'(a,d21.10)') trim(caller_subname)//'accum_renorm_factor = ', accum_renorm_factor
-          write(logunit,'(a,d21.10)') trim(caller_subname)//'ablat_renorm_factor = ', ablat_renorm_factor
+          write(logunit,'(a,d28.17)') trim(caller_subname)//'accum_renorm_factor = ', accum_renorm_factor
+          write(logunit,'(a,d28.17)') trim(caller_subname)//'ablat_renorm_factor = ', ablat_renorm_factor
        endif
     endif
 
--- a/mediator/med_phases_prep_glc_mod.F90
+++ b/mediator/med_phases_prep_glc_mod.F90
@@ -51,6 +51,7 @@ module med_phases_prep_glc_mod
   use wtracers_mod          , only : wtracers_present, wtracers_get_num_tracers, WTRACERS_SUFFIX
   use perf_mod              , only : t_startf, t_stopf
   use shr_log_mod           , only : shr_log_error
+  use shr_reprosum_mod      , only : shr_reprosum_calc
 
   implicit none
   private
@@ -63,8 +64,9 @@ module med_phases_prep_glc_mod
 
   private :: med_phases_prep_glc_map_lnd2glc
   private :: med_phases_prep_glc_renormalize_smb
-  private :: renormalize_smb_accumulate_sums_l
-  private :: renormalize_smb_accumulate_sums_g
+  private :: renormalize_smb_compute_summands_l
+  private :: renormalize_smb_compute_summands_g
+  private :: renormalize_smb_global_sums
   private :: renormalize_smb_do_renormalization
 
   ! -----------------
@@ -80,6 +82,9 @@ module med_phases_prep_glc_mod
   ! Should be set to true for 2-way coupled runs with evolving ice sheets.
   ! Does not need to be true for 1-way coupling.
   logical :: smb_renormalize
+
+  ! if true, compute global sums in a manner that is independent of processor count
+  logical :: bfbflag
 
   type(ESMF_FieldBundle), public :: FBlndAccum2glc_l
   integer               , public :: lndAccum2glc_cnt
@@ -151,6 +156,7 @@ contains
     type(ESMF_Mesh)     :: mesh_o
     type(ESMF_Field)    :: lfield
     character(len=CS)   :: glc_renormalize_smb
+    character(len=CS)   :: cvalue
     integer             :: ungriddedUBound_output(1) ! currently the size must equal 1 for rank 2 fieldds
     character(len=*),parameter  :: subname=' (med_phases_prep_glc_init) '
     !---------------------------------------
@@ -290,6 +296,14 @@ contains
        end select
        if (maintask) then
           write(logunit,'(a,l4)') trim(subname)//' smb_renormalize is ',smb_renormalize
+       end if
+
+       ! Determine whether to compute global sums in a processor-count-independent manner
+       call NUOPC_CompAttributeGet(gcomp, name='bfbflag', value=cvalue, rc=rc)
+       if (chkerr(rc,__LINE__,u_FILE_u)) return
+       read(cvalue,*) bfbflag
+       if (maintask) then
+          write(logunit,'(a,l4)') trim(subname)//' bfbflag is ',bfbflag
        end if
 
        if (smb_renormalize) then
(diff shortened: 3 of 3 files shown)


billsacks:
I want to do a little more testing to verify that this is all working as expected... I'll post here when that's complete.

Katetc:
I just wanted to start my review by saying this line in the PR Summary:

> Potential roundoff-level changes for tests that turn on BFBFLAG (ERP, PEM, PET, SEQ)

Feels kind of ironic. I understand that you mean the initial answers, not the restarted (repro_sum'd) answers, but maybe it's worth qualifying which answers are BFB and which are roundoff-level changed.

cime_config/config_component.xml
```diff
@@ -1107,6 +1107,20 @@
...
 
+  <entry id="BFBFLAG">
+    <type>logical</type>
+    <valid_values>TRUE,FALSE</valid_values>
+    <default_value>FALSE</default_value>
+    <group>run_flags</group>
+    <file>env_run.xml</file>
+    <desc>
+      If TRUE, perform some calculations (in CMEPS and elsewhere) such as global sums in a
+      manner that is bit-for-bit independent of processor count.
```
Katetc: "If TRUE, perform some calculations (in CMEPS and elsewhere) such as global sums using a less efficient algorithm that is independent of processor count. This allows for Bit-For-Bit (BFB) restarts after changing PE layouts, for many compsets. However, setting this flag does NOT guarantee that all calculations throughout all model configurations are BFB independent of processor count."

cime_config/namelist_definition_drv.xml
```diff
@@ -2676,6 +2676,25 @@
...
+         sums. -->
+  <entry id="bfbflag" modify_via_xml="BFBFLAG">
+    <type>logical</type>
+    <category>reprosum</category>
+    <group>ALLCOMP_attributes</group>
+    <desc>
+      If true, perform some calculations (in CMEPS and elsewhere) such as global sums in a
+      manner that is bit-for-bit independent of processor count.
+      (This does NOT guarantee that all calculations throughout the model are bit-for-bit
+      independent of processor count.)
```
Katetc: I find it a little weird that this text is in two places, but whatever. If you like my suggestion above, you could update this one as well.

billsacks:
> I just wanted to start my review by saying this line in the PR Summary:
> 
> > Potential roundoff-level changes for tests that turn on BFBFLAG (ERP, PEM, PET, SEQ)
> 
> Feels kind of ironic. I understand that you mean the initial answers, not the restarted (repro_sum'd) answers, but maybe it's worth qualifying which answers are BFB and which are roundoff-level changed.

Point taken. I tried adjusting this in the top-level PR description. See if this reads better now.

mediator/med_global_sums_mod.F90
```diff
@@ -0,0 +1,161 @@
...
+
+    ! input/output variables
+    type(ESMF_GridComp), intent(in)  :: gcomp
+    real(r8)           , intent(in)  :: local_summands(:)
+    real(r8)           , intent(out) :: global_sum
+    integer            , intent(out) :: rc
+
+    ! local variables
+    ! Note that summands_2d is deliberately allocatable, rather than being an automatic array or
+    ! an inline reshape of local_summands: those would give an array temporary, which some
```
Katetc: give an array temporary... memory?

mediator/med_global_sums_mod.F90
```diff
@@ -0,0 +1,161 @@
...
+    ! compilers place on the stack, and local_summands can be large (e.g., of order the number of
+    ! local grid cells).
+    real(r8), allocatable :: summands_2d(:,:)
+    real(r8)              :: global_sums(1)
+    !---------------------------------------------------------------
+
+    rc = ESMF_SUCCESS
+
+    allocate(summands_2d(size(local_summands), 1))
```
Katetc, line 149: Ok, I was feeling like a check was needed here, and I figured out what I was thinking is adding a return stat in the allocate call and checking to make sure it was successful. Like: 
`allocate(summands_2d(size(local_summands), 1), stat=ierr)

  if (ierr /= 0) then
        errstring='Error allocating summands_2d'`

mediator/med_phases_post_rof_mod.F90
```diff
@@ -358,36 +361,36 @@ subroutine med_phases_post_rof_remove_negative_runoff(gcomp, field_name, rc)
     call fldbun_getdata1d(FBrof_r, trim(field_name), runoff_flux, rc=rc)
     if (ChkErr(rc,__LINE__,u_FILE_u)) return
 
-    local_positive(1) = 0.0_r8
-    local_negative(1) = 0.0_r8
+    allocate(local_positive(size(runoff_flux)))
+    allocate(local_negative(size(runoff_flux)))
```
Katetc, line 365: Maybe a check for error status on the allocate?

mediator/med_phases_post_rof_mod.F90
```diff
@@ -358,36 +361,36 @@ subroutine med_phases_post_rof_remove_negative_runoff(gcomp, field_name, rc)
...
-    call ESMF_VMAllreduce(vm, senddata=local_positive, recvdata=global_positive, count=1, &
-         reduceflag=ESMF_REDUCE_SUM, rc=rc)
+    call med_global_sums(gcomp, local_negative, global_negative, rc)
     if (ChkErr(rc,__LINE__,u_FILE_u)) return
-    call ESMF_VMAllreduce(vm, senddata=local_negative, recvdata=global_negative, count=1, &
-         reduceflag=ESMF_REDUCE_SUM, rc=rc)
-    if (ChkErr(rc,__LINE__,u_FILE_u)) return
-    global_sum = global_positive(1) + global_negative(1)
+    deallocate(local_positive)
+    deallocate(local_negative)
```
Katetc, line 381: Maybe a check to make sure it's allocated before deallocating?

mediator/med_phases_post_rof_mod.F90
```diff
@@ -417,26 +420,28 @@ subroutine med_phases_post_rof_remove_negative_runoff(gcomp, field_name, rc)
 
     if (dbug_flag > dbug_threshold) then
       ! Recompute positives, negatives and total sum for output diagnostic purposes
-      local_positive_final(1) = 0.0_r8
-      local_negative_final(1) = 0.0_r8
+      allocate(local_positive_final(size(runoff_flux)))
+      allocate(local_negative_final(size(runoff_flux)))
```
Katetc, line 424: Safety checks on these allocates?

mediator/med_phases_post_rof_mod.F90
```diff
@@ -417,26 +420,28 @@ subroutine med_phases_post_rof_remove_negative_runoff(gcomp, field_name, rc)
...
-           reduceflag=ESMF_REDUCE_SUM, rc=rc)
+      call med_global_sums(gcomp, local_positive_final, global_positive_final, rc)
       if (ChkErr(rc,__LINE__,u_FILE_u)) return
-      call ESMF_VMAllreduce(vm, senddata=local_negative_final, recvdata=global_negative_final, count=1, &
-           reduceflag=ESMF_REDUCE_SUM, rc=rc)
+      call med_global_sums(gcomp, local_negative_final, global_negative_final, rc)
       if (ChkErr(rc,__LINE__,u_FILE_u)) return
-      global_sum_final = global_positive_final(1) + global_negative_final(1)
+      deallocate(local_positive_final)
+      deallocate(local_negative_final)
```
Katetc, line 439: And checks here as well

cime_config/config_component.xml
billsacks: I have partially reworded this based on your suggestion (e15174344f2fd105bfbcd3395a5a0d15c65aa0c0). I left out the mention of "restarts" because I felt that was a bit confusing: Even though this shows up in restart tests (ERP) it is not directly related to restarts. See if you're happy with the new wording in that commit.

mediator/med_phases_post_rof_mod.F90
```diff
@@ -417,26 +420,28 @@ subroutine med_phases_post_rof_remove_negative_runoff(gcomp, field_name, rc)
...
+      global_sum_final = global_positive_final + global_negative_final
       if (maintask) then
```
Katetc, line 441: So, outside the scope of these changes, but the writes above included "dbug_flag > dbug_threshold" in this check and I think that might be good here to reduce log output.

mediator/med_phases_prep_glc_mod.F90
```diff
@@ -1226,40 +1232,40 @@ subroutine med_phases_prep_glc_renormalize_smb(gcomp, ns, rc)
        effective_area_l(n) = min(lndfrac(n), icemask_l(n)) * is_local%wrap%mesh_info(complnd)%areas(n)
     end do
 
-    ! determine accumulation and ablation sums for qice on the land grid
+    ! determine accumulation and ablation summands for qice on the land grid
+    allocate(local_accum_lnd(size(lndfrac), 1+num_wtracers))
+    allocate(local_ablat_lnd(size(lndfrac), 1+num_wtracers))
```
Katetc, line 1237: I'm going to say "Safety checks?" here as well, but I'm starting to doubt myself.

mediator/med_phases_prep_glc_mod.F90
```diff
@@ -1226,40 +1232,40 @@ subroutine med_phases_prep_glc_renormalize_smb(gcomp, ns, rc)
...
-    call ESMF_VMAllreduce(vm, senddata=local_accum_lnd, recvdata=global_accum_lnd, count=1+num_wtracers, &
-         reduceflag=ESMF_REDUCE_SUM, rc=rc)
+    call med_global_sums(gcomp, local_accum_lnd, global_accum_lnd, rc)
     if (ChkErr(rc,__LINE__,u_FILE_u)) return
-    call ESMF_VMAllreduce(vm, senddata=local_ablat_lnd, recvdata=global_ablat_lnd, count=1+num_wtracers, &
-         reduceflag=ESMF_REDUCE_SUM, rc=rc)
+    call med_global_sums(gcomp, local_ablat_lnd, global_ablat_lnd, rc)
     if (ChkErr(rc,__LINE__,u_FILE_u)) return
+    deallocate(local_accum_lnd)
+    deallocate(local_ablat_lnd)
```
Katetc, line 1261: Check to see if allocated first

Katetc (approved):
There's nothing here that is a show stopper. I've added some wording suggestions in a few places, and just thought that adding some checks around allocate and deallocate calls are usually a good idea. But, we are in a time crunch, so feel free to just mark those resolved if it's not worth making changes.

cime_config/namelist_definition_drv.xml
billsacks: > I find it a little weird that this text is in two places, but whatever. If you like my suggestion above, you could update this one as well.

Yeah, I hear you. Since both of them will appear in the documentation in separate places - depending on whether you are looking at the documentation of xml variables or namelist variables - I thought it was worth documenting it in both places. I suppose I could just say, "see documentation of BFBFLAG xml variable". Would you prefer that?

mediator/med_global_sums_mod.F90
billsacks: > give an array temporary... memory?

You're right - that wording is bad. I changed it to "create an array temporary", which is what it means.

I was on the fence about whether to keep this comment at all. Claude added it when I asked Claude to change the implementation to avoid creating an array temporary. I almost deleted it, feeling it was unnecessary, but then decided that it provides a little value. Looking back on it, I am tempted to delete it. Let me know if you have preferences.

mediator/med_global_sums_mod.F90
billsacks: I realize that feelings differ across CSEG on this point (with CAM apparently preferring a stat check). I personally prefer to *not* have an explicit check of the stat of an allocate: if you don't query the stat, then the code will abort if there's an error in allocation; if you do query the stat, then it's up to you to handle the error. This leaves open the possibility that you'll make an error in the error handling itself, accidentally letting an allocation error slip through without aborting. I think I've seen this kind of bug in the error handling of an allocate in a recent code review. I realize that catching the error can allow for a more informative error message, but my feeling is that it's not worth the extra code and the possibility of inserting a bug in the error-handling code.

I'm open to having my mind changed on this point, but I think this is a bigger discussion that should be taken outside of this PR review: the vast majority of allocates in CMEPS do *not* check the stat, so I prefer to stay consistent with the pre-existing pattern until we have a discussion where the consensus is that we want to add this pattern.

mediator/med_phases_post_rof_mod.F90
billsacks: > Maybe a check to make sure it's allocated before deallocating?

This is another conversation that I think should be taken outside this PR review if you think it's important: From my quick look, it looks like we rarely check the allocated status of arrays before deallocating in CMEPS, and it seems unnecessary here because they are allocated unconditionally. (I have a feeling that the places that do check the allocation status are handling situations where the array may truly be unallocated at that point - for example, I see this in a "clean" routine that I think I may have written, which is handling the possibility that "clean" is called - e.g., from a unit test - despite the array not having been allocated.) So this would be a broader discussion on the preferred pattern in general.

Actually, I think it would be safe to completely remove these deallocate statements, because since Fortran 2003 allocatables are automatically deallocated when they go out of scope. I thought about doing that, and would personally prefer removing them completely over adding a conditional check around the deallocate.

mediator/med_phases_post_rof_mod.F90
billsacks: This whole block of code is already inside a conditional on `dbug_flag > dbug_threshold`.

billsacks:
@Katetc - thank you very, very much for your quick turnaround of this! I have addressed some of your comments with improved wording in comments / descriptions (thanks for those suggestions!). I replied to your comments about adding checks around allocates / deallocates. I think I have addressed everything at this point, at least if you're happy with how I addressed them.

I'm still running some final testing... derecho's slow queues have been slowing me down today.

billsacks:
I have done some additional testing, which I have documented in the top-level comment of this PR.

cime_config/config_component.xml
Katetc: I like it!

cime_config/namelist_definition_drv.xml
Katetc: I think it's fine to have the same text in both places. Redundancy should generally be avoided, but it's not the end of the world. :D

mediator/med_global_sums_mod.F90
Katetc: I support deleting this comment. I'm glad you are thinking about these things, but you don't have to defend every piece of code from potential criticism.

mediator/med_global_sums_mod.F90
Katetc: Hi Bill, no worries. I think the fact that this is in-line with CMEPS standards is good enough. I've had too many reviews where CAM SEs (no need to name names!) would flag all of mine, so I just want to make sure we at least consider the choice.

mediator/med_phases_post_rof_mod.F90
Katetc: Hmm, TIL. Well, most of my Fortran has this check in it, both CISM and CAM. Mostly in case the code changes in a way that could result in the array being unallocated when the deallocate is called. I agree it's not super necessary here.

billsacks: > Hmm, TIL. Well, most of my Fortran has this check in it, both CISM and CAM. Mostly in case the code changes in a way that could result in the array being unallocated when the deallocate is called. I agree it's not super necessary here.

Fair enough. I see that rationale. Personally, I think I'd look at that and say, "why is this only conditionally deallocated? what is the situation where it wouldn't have been allocated?" But I certainly see your perspective on this, too, and it's a good one.

billsacks:
@Katetc gave me the go-ahead to merge....

